PR #17, opened 2020-05-05
PLATUI-441: Add currency input component from hmrc-frontend

matthewmascord:
Allow additional set of test fixtures (in addition to those published within hmrc-frontend)

It is worth noting that this is the first time play-frontend-hmrc uses play-frontend-govuk. This is because hmrcCurrencyInput uses govukLabel and govukHint in the same way that the hmrc-frontend Nunjucks components do.


natcarey:
Actually, the new tests won't add anything because the documentation for this isn't ready yet.

PR #20, opened 2020-07-16
Upgrade to 1.15.1

natcarey:
This includes one interface change (NavigationItem) and one overloaded constructor (NotificationBadge).


src/test/scala/uk/gov/hmrc/hmrcfrontend/views/viewmodels/notificationbadge/NotificationBadgeSpec.scala
```diff
@@ -16,7 +16,39 @@
...
+    "accept strings" in {
+      val testData = NotificationBadge("3")
+
+      val json = Json.toJson[NotificationBadge](testData)
+
+      json.as[NotificationBadge] should be(
+        NotificationBadge("3"))
+    }
+    "accept integers" in {
+      val testData = NotificationBadge(1)
```
matthewmascord, line 34: I'm not sure whether this test really proves anything, given the integer 1 will be implicitly converted to a string.

project/LibDependencies.scala
```diff
@@ -34,7 +34,7 @@ object LibDependencies {
     },
     play25 = {
       val compile = Seq(
-        "uk.gov.hmrc" %% "play-frontend-govuk" % "0.47.0-play-25"
+        "uk.gov.hmrc" %% "play-frontend-govuk" % "0.49.0-play-25"
```
JoPintoPaul, line 37: Shall we pull the version out into a string, like we do with hmrcFrontendVersion?

src/main/play-26/twirl/uk/gov/hmrc/hmrcfrontend/views/components/hmrcHeader.scala.html
```diff
@@ -94,20 +94,20 @@
...
-      <button type="button" role="button" class="govuk-header__menu-button js-header-toggle"
+      <button type="button" class="govuk-header__menu-button govuk-js-header-toggle"
           aria-controls="navigation"
           aria-label="Show or hide Top Level Navigation">Menu
       </button>
       <nav>
         <ul id="navigation"
           class="@toClasses("govuk-header__navigation ", navigationClasses)"
           aria-label="Top Level Navigation">
-          @for(item <- navigation.getOrElse(Seq.empty).filter(item => item.href.nonEmpty && item.text.nonEmpty)) {
```
matthewmascord, line 105: I think this is the bit that is making the integration test fail. Previously we were ignoring any entries that had an empty href or text. We're now including navigation items where href is empty.

matthewmascord: The filter probably needs changing to something like item => item.href.nonEmpty && (item.text.nonEmpty || item.content.nonEmpty)

matthewmascord: Worth adding a specific unit test or additional-fixture for this case.

PR #25, opened 2020-10-16
PLATUI-606: Add Welsh messages to hmrcFooter

matthewmascord:


src/main/play-26/twirl/uk/gov/hmrc/hmrcfrontend/views/components/hmrcFooter.scala.html
```diff
@@ -14,11 +14,69 @@
...
-@import uk.gov.hmrc.govukfrontend.views.html.components.GovukFooter
 @import uk.gov.hmrc.govukfrontend.views.viewmodels.footer.{Footer, Meta}
 @import uk.gov.hmrc.hmrcfrontend.views.config.HmrcFooterItems
 
-@this(govukFooter: GovukFooter, hmrcFooterItems: HmrcFooterItems)
+@this(hmrcFooterItems: HmrcFooterItems)
 
 @()(implicit messages: Messages, request: RequestHeader)
-@GovukFooter(new Footer(meta = Some(Meta(items = Some(hmrcFooterItems.get)))))
+@defining(Meta(items = Some(hmrcFooterItems.get))) { meta =>
```
natcarey, line 23: (as discussed in meetings) This is a change from the normal approach, I think we should build this into `hmrc-frontend` first and then port it over to keep consistency with the prototype kit.

PR #40, opened 2021-01-07
PLATUI-795: Deleted all Play 2.5 specific code and updated HMRC heade…

JoPintoPaul:
…r year to 2021


src/main/scala/uk/gov/hmrc/hmrcfrontend/views/viewmodels/JsonDefaultValueFormatter.scala
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright 2020 HM Revenue & Customs
+ * Copyright 2021 HM Revenue & Customs
```
matthewmascord, line 2: Should we remove this class as suggested in the FIXME?

JoPintoPaul: @matthewmascord Have created a Jira ticket to do this separately.

matthewmascord (approved):

PR #43, opened 2021-01-20
PLATUI-848: Uplift SBT to 1.4.6

matthewmascord:


src/test/scala/uk/gov/hmrc/hmrcfrontend/views/helpers/hmrcReportTechnicalIssueHelperSpec.scala
```diff
@@ -32,48 +33,99 @@ class hmrcReportTechnicalIssueHelperSpec
...
-      .build()
-
   def buildApp(properties: Map[String, String] = Map.empty): Application =
     new GuiceApplicationBuilder()
       .configure(Map("play.allowGlobalApplication" -> "true") ++ properties)
       .build()
 
-  "HmrcReportTechnicalIssueHelper" should {
-    "render link when serviceId is set" in {
+  "HmrcReportTechnicalIssueHelperDependencyInjection" should {
```
JoPintoPaul, line 41: Is that the actual name of the component? Might suggest "HmrcReportTechnicalIssueHelper via dependency injection", or changing it so it doesn't sound like the actual component name. At the moment, this reads like there is a component `HmrcReportTechnicalIssueHelperDependencyInjection`.

matthewmascord: Agreed, this was copied over from a specific set of DI tests. I think it should just be the name of the component.

JoPintoPaul (approved):

PR #62, opened 2021-03-19
PLATUI-909: Integrate play language

matthewmascord:


src/test/scala/uk/gov/hmrc/hmrcfrontend/controllers/LanguageControllerSpec.scala
```diff
@@ -0,0 +1,88 @@
...
+import org.scalatest.wordspec.AnyWordSpec
+import org.scalatestplus.play.guice.GuiceOneAppPerSuite
+import play.api.Application
+import play.api.http.{HeaderNames, Status}
+import play.api.inject.guice.GuiceApplicationBuilder
+import play.api.mvc.AnyContentAsEmpty
+import play.api.test.FakeRequest
+import play.api.test.Helpers._
+
+import scala.collection.immutable.List
```
JoPintoPaul, line 29: Tested locally, seems that this `immutable` version of `List` isn't required.

JoPintoPaul:
Looks good, just one code suggestion.

Also, is it worth adding something to the backlog to look at inlining the `play-language` library, once we are able to deprecate the standalone one?

JoPintoPaul (approved):

PR #72, opened 2021-05-10
PLATUI-442: add hmrcAddToAList component

oscarduignan:
I've got a few questions about how to handle a dependency that I've left in as TODOs

https://github.com/hmrc/play-frontend-hmrc/compare/PLATUI-442_update-hmrc-frontend-version...PLATUI-442_add-to-a-list-component?expand=1#diff-4a7fb8b748eb4a6d5c48ad1936a81416303c12047f46c39ecf6c707bf775c15cR25-R27

but before trying to come to some consensus on those it's probably better to look at how I'm using the helper in question within the twirl template https://github.com/hmrc/play-frontend-hmrc/compare/PLATUI-442_update-hmrc-frontend-version...PLATUI-442_add-to-a-list-component?expand=1#diff-939d572314852a3ec07c66afe543de5e7fc28258227eade12bb183e197c88516

### DIFF
--- a/src/main/scala/uk/gov/hmrc/hmrcfrontend/views/html/components/package.scala
+++ b/src/main/scala/uk/gov/hmrc/hmrcfrontend/views/html/components/package.scala
@@ -17,7 +17,7 @@
 package uk.gov.hmrc.hmrcfrontend.views
 package html
 
-import uk.gov.hmrc.govukfrontend.views.html.components.{GovukErrorMessage, GovukHint, GovukLabel, GovukPhaseBanner, GovukTextarea}
+import uk.gov.hmrc.govukfrontend.views.html.components._
 
 package object components extends Utils with Aliases {
 
@@ -81,5 +81,9 @@ package object components extends Utils with Aliases {
   type HmrcCharacterCount = hmrcCharacterCount
   @deprecated(message = "Use DI", since = "Play 2.6")
   lazy val HmrcCharacterCount = new hmrcCharacterCount(GovukTextarea, GovukHint)
+
+  type HmrcAddToAList = hmrcAddToAList
+  @deprecated(message = "Use DI", since = "Play 2.6")
+  lazy val HmrcAddToAList = new hmrcAddToAList(GovukRadios, GovukButton)
 
 }
--- a/src/main/twirl/uk/gov/hmrc/hmrcfrontend/views/components/hmrcAddToAList.scala.html
+++ b/src/main/twirl/uk/gov/hmrc/hmrcfrontend/views/components/hmrcAddToAList.scala.html
@@ -1,0 +1,138 @@
+@*
+ * Copyright 2021 HM Revenue & Customs
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *@
+
+@import uk.gov.hmrc.govukfrontend.views.Aliases.Hint
+@import uk.gov.hmrc.govukfrontend.views.html.components.{GovukButton, GovukRadios}
+@import uk.gov.hmrc.govukfrontend.views.viewmodels.button.Button
+@import uk.gov.hmrc.govukfrontend.views.viewmodels.content.Text
+@import uk.gov.hmrc.govukfrontend.views.viewmodels.fieldset.{Fieldset, Legend}
+@import uk.gov.hmrc.govukfrontend.views.viewmodels.radios.{RadioItem, Radios}
+@import uk.gov.hmrc.hmrcfrontend.views.viewmodels.language.{Cy, En}
+@import uk.gov.hmrc.hmrcfrontend.views.viewmodels.addtoalist.AddToAList
+@import uk.gov.hmrc.hmrcfrontend.views.viewmodels.addtoalist.ItemType
+
+@this(govukRadios: GovukRadios, govukButton: GovukButton)
+
+@(params: AddToAList)(implicit messages: Messages)
+@import params._
+
+@defining(if(params.language == Cy || messages.lang.code == "cy") Cy else En) { language =>
+
+  @defining(
+    itemType.getOrElse(language match {
+      case Cy => ItemType(
+        singular = "eitem",
+        plural = "o eitemau")
+      case _ => ItemType(
+        singular = "item",
+        plural = "items"
+      )
+    })
+  ) { itemType =>
+
+    <h1 class="govuk-heading-xl">
+      @(language, itemList.length) match {
+        case (Cy, 0) => { Nid ydych wedi ychwanegu unrhyw eitemau }
+        case (_, 0) => { You have not added any items }
+        case (Cy, 1) => { Rydych wedi ychwanegu 1 @itemType.singular}
+        case (_, 1) => { You have added 1 @itemType.singular}
+        case (Cy, n) => { Rydych wedi ychwanegu @n @itemType.plural}
+        case (_, n) => { You have added @n @itemType.plural}
+      }
+    </h1>
+
+    <div class="govuk-form-group">
+      <dl class="hmrc-add-to-a-list hmrc-add-to-a-list--@itemSize.suffix">
+      @for(item <- itemList) {
+        <div class="hmrc-add-to-a-list__contents">
+          <dt class="hmrc-add-to-a-list__identifier hmrc-add-to-a-list__identifier--light">
+          @item.name
+          </dt>
+          <dd class="hmrc-add-to-a-list__change">
+            <a class="govuk-link" href="@item.changeUrl">
+              <span aria-hidden="true">@language match {
+                case Cy => { Newid }
+                case _ => { Change }
+              }</span>
+              <span class="govuk-visually-hidden">@language match {
+                case Cy => { Newid @item.name}
+                case _ => { Change @item.name}
+              }</span>
+            </a>
+          </dd>
+          <dd class="hmrc-add-to-a-list__remove">
+            <a class="govuk-link" href="@item.removeUrl">
+              <span aria-hidden="true">@language match {
+                case Cy => { Dileu }
+                case _ => { Remove }
+              }</span>
+              <span class="govuk-visually-hidden">@language match {
+                case Cy => { Dileu’r @item.name o’r rhestr }
+                case _ => { Remove @item.name from the list }
+              }</span>
+            </a>
+          </dd>
+        </div>
+      }
+      </dl>
+    </div>
+    <form method="post" action="@formAction">
+      @govukRadios(Radios(
+        fieldset = Some(Fieldset(
+          legend = Some(Legend(
+            content = Text(language match {
+              case Cy => s"Oes angen i chi ychwanegu ${itemType.singular} arall?"
+              case _ => s"Do you need to add another ${itemType.singular}?"
+            }),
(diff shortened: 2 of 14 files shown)


src/main/scala/uk/gov/hmrc/hmrcfrontend/views/helpers/MessageHelpers.scala
```diff
@@ -0,0 +1,39 @@
...
+
+package uk.gov.hmrc.hmrcfrontend.views.helpers
+
+import play.api.i18n.{Lang, Messages, MessagesApi, MessagesImpl}
+
+object MessageHelpers {
+  def messagesForLanguage(
+    code: Option[String] = None
+  )(implicit messages: Messages, messagesApi: MessagesApi): Messages =
+    // TODO: PR question: should I make this a class and @inject messagesApi and pass helper via DI?
```
oscarduignan, line 25: I feel like I prefer making it a class and injecting messages Api rather than asking for it as an implicit, but I hadn't found `Play.current.injector.instanceOf[]` yet to allow for https://github.com/hmrc/play-frontend-hmrc/blob/master/src/main/scala/uk/gov/hmrc/hmrcfrontend/views/html/components/package.scala

src/main/scala/uk/gov/hmrc/hmrcfrontend/views/helpers/MessageHelpers.scala
```diff
@@ -0,0 +1,39 @@
...
+    // TODO: PR question: should I require lang supplied be supported / print a warning / solely rely on messagesApi fallback behaviour?
```
oscarduignan, line 26: I probably favour raising an exception if the lang isn't in `play.api.Langs.availables` but didn't implement because with current implementation it would need another implicit, but if I refactor and to @inject instead then it's easy to add

src/main/scala/uk/gov/hmrc/hmrcfrontend/views/helpers/MessageHelpers.scala
```diff
@@ -0,0 +1,39 @@
...
+    // TODO: PR question: should parameter be Option[Lang]?
```
oscarduignan, line 27: wasn't sure if we wanted to keep Lang out of twirl templates - so just made it a string to start with

src/main/twirl/uk/gov/hmrc/hmrcfrontend/views/components/hmrcAddToAList.scala.html
```diff
@@ -0,0 +1,104 @@
...
+                        "addToAList.addAnother",
+                        itemType.map { itemType => itemType.singular }.getOrElse {
+                            messages(pluralMessageKey("addToAList.defaultItemType", 1))
+                        }
+                    )),
+                    classes = "govuk-fieldset__legend govuk-fieldset__legend--m",
+                    isPageHeading = false
+                ))
+            )),
+            hint = Some(Hint(content = Text(hintText.getOrElse("")))),
```
oscarduignan, line 85: - [ ] going to raise a possible bug because this is always empty

src/main/twirl/uk/gov/hmrc/hmrcfrontend/views/components/hmrcAddToAList.scala.html
```diff
@@ -0,0 +1,104 @@
...
+        @govukRadios(Radios(
+            fieldset = Some(Fieldset(
+                legend = Some(Legend(
+                    content = Text(messages(
+                        "addToAList.addAnother",
+                        itemType.map { itemType => itemType.singular }.getOrElse {
+                            messages(pluralMessageKey("addToAList.defaultItemType", 1))
+                        }
+                    )),
+                    classes = "govuk-fieldset__legend govuk-fieldset__legend--m",
```
oscarduignan, line 81: - [ ] going to raise a bug seperately, leads to element having 2 `govuk-fieldset__legend` classes

src/main/twirl/uk/gov/hmrc/hmrcfrontend/views/components/hmrcAddToAList.scala.html
```diff
@@ -0,0 +1,104 @@
...
+                    <a class="govuk-link" href="@item.removeUrl">
+                        <span aria-hidden="true">@messages("addToAList.remove.displayed")</span>
+                        <span class="govuk-visually-hidden">@messages("addToAList.remove.announced", item.name)</span>
+                    </a>
+                </dd>
+            </div>
+        }
+        </dl>
+    </div>
+    <form method="post" action="@formAction">
```
oscarduignan, line 71: - [ ] going to raise seperately, or at least ask, should this be a post? Maybe the pattern is POST to create empty thing -> then edit it; I want to check because I'm more used to GET form to create new entity -> POST to create

src/main/twirl/uk/gov/hmrc/hmrcfrontend/views/components/hmrcAddToAList.scala.html
```diff
@@ -0,0 +1,104 @@
+@*
```
oscarduignan, line 1: just realised that I forgot to fix the indentation on this file. Intellij defaulted to 4 spaces for twirl templates and I never remembered to adjust the setting, I'll push an update that reformats this file after lunch 👍

src/main/resources/messages
```diff
@@ -15,3 +15,18 @@ footer.contact.text = Contact
 footer.contact.url = https://www.gov.uk/government/organisations/hm-revenue-customs/contact
 footer.welshHelp.text = Rhestr o Wasanaethau Cymraeg
 footer.welshHelp.url = https://www.gov.uk/cymraeg
+
+addToAList.heading.0 = You have not added any items
```
matthewmascord, line 19: I'd be in favour of adding messages here to be more idiomatic. However, I feel we should hold off on this for now until we can implement a solution that will be consistent for all our components. What do you think?

src/main/twirl/uk/gov/hmrc/hmrcfrontend/views/components/hmrcAddToAList.scala.html
matthewmascord: It's a state-changing request, so I would expect the method to be a POST and to protect against CSRF. I'm probably missing some context in your question though :-)

src/main/twirl/uk/gov/hmrc/hmrcfrontend/views/components/hmrcAddToAList.scala.html
```diff
@@ -0,0 +1,104 @@
...
+@import uk.gov.hmrc.govukfrontend.views.viewmodels.button.Button
+@import uk.gov.hmrc.govukfrontend.views.viewmodels.content.Text
+@import uk.gov.hmrc.govukfrontend.views.viewmodels.fieldset.{Fieldset, Legend}
+@import uk.gov.hmrc.govukfrontend.views.viewmodels.radios.{RadioItem, Radios}
+@import uk.gov.hmrc.hmrcfrontend.views.helpers.MessageHelpers.{messagesForLanguage, pluralMessageKey}
+@import uk.gov.hmrc.hmrcfrontend.views.viewmodels.addtoalist.AddToAList
+
+@this(govukRadios: GovukRadios, govukButton: GovukButton)
+
+@(params: AddToAList)(implicit messages: Messages, messagesApi: MessagesApi)
```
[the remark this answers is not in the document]
oscarduignan: With that one my thought was that if we're already requiring an implicit messages api we might as well use the messages rather than just the language https://github.com/hmrc/play-frontend-hmrc/blob/2ba82c75f6284ece52a3a034996dbdd711058f0a/src/main/twirl/uk/gov/hmrc/hmrcfrontend/views/components/hmrcCharacterCount.scala.html#L25 for this if I @inject messagesApi into the messagesForLanguages and make that a DI dependency then they wouldn't need to pass anything other than the implicit messages like the other components

src/main/twirl/uk/gov/hmrc/hmrcfrontend/views/components/hmrcAddToAList.scala.html
oscarduignan: my thought on this was is submitting this form actually going to create another thing? Or will it just take you to the form that you would then submit to add another (so the state change would be submitting that next form), feels like here we've not got an inline form for creating an "item" so much as we have a question about if you'd like to add another. I should probably see if I can find an example of it in use and talk to design system team - probably that would clear up intended use here

src/main/twirl/uk/gov/hmrc/hmrcfrontend/views/components/hmrcAddToAList.scala.html
oscarduignan: yup, sorry I wasn't clear I'll edit it, I meant raise it against hmrc-frontend - I'm just matching the nunjucks template here

src/main/twirl/uk/gov/hmrc/hmrcfrontend/views/components/hmrcAddToAList.scala.html
oscarduignan: yup, same as above this is a bug with hmrc-frontend, all of these tasks are questions / tickets for hmrc-frontend

src/main/twirl/uk/gov/hmrc/hmrcfrontend/views/components/hmrcAddToAList.scala.html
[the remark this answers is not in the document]
oscarduignan: I'll update this to quickly to match what I was thinking as an example, but if that still isn't quite right after review then happy to simplify, won't be much work to try both first I think 👍

src/main/scala/uk/gov/hmrc/hmrcfrontend/views/html/components/package.scala
```diff
@@ -82,4 +83,8 @@ package object components extends Utils with Aliases {
   @deprecated(message = "Use DI", since = "Play 2.6")
   lazy val HmrcCharacterCount = new hmrcCharacterCount(GovukTextarea, GovukHint)
 
+  type HmrcAddToAList = hmrcAddToAList
+  @deprecated(message = "Use DI", since = "Play 2.6")
+  lazy val HmrcAddToAList = Play.current.injector.instanceOf[HmrcAddToAList]
```
oscarduignan, line 88: This doesn't work, but I'm not quite sure what the important bit I've missed that's different from https://github.com/hmrc/play-frontend-hmrc/blob/master/src/main/scala/uk/gov/hmrc/hmrcfrontend/views/html/deprecated/deprecatedPlay26Helpers.scala is 🤨 . @JoPintoPaul and @matthewmascord would either of you be able to help me understand what's gone wrong here? I'm unsure why this won't build on CI but does locally, I'm missing something crucial

oscarduignan: I understand the issue now that it's using deprecated functionality that isn't compatible with 2.8, not sure if there's any salvageable alternative here or I've hit a limitation I hadn't come across yet with these static components still around and I guess if so will need to change the implementation to still be able to support this

matthewmascord: Play.current isn't a think in 2.8. Can you instantiate with new as with the ones above?

matthewmascord (approved):

src/main/twirl/uk/gov/hmrc/hmrcfrontend/views/components/hmrcAddToAList.scala.html
natcarey: I agree, I hadn't noticed before but at this stage you're essentially specifying which page you want to see next so a GET would be perfectly suitable.

I think there's an unspoken standard of these kind of actions being POST requests (e.g. https://www.tax.service.gov.uk/pay).  I'd suggest doing this one "properly" and use it as an example - safe and idempotent actions are GET requests, unsafe or non-idempotent actions are POST request.

natcarey (approved):

PR #77, opened 2021-05-24
PLATUI-1123 Add timeline component

natcarey:


src/main/scala/uk/gov/hmrc/hmrcfrontend/views/Aliases.scala
```diff
@@ -101,6 +101,9 @@ trait Aliases {
 
   type AddToAList = viewmodels.addtoalist.AddToAList
   val AddToAList = viewmodels.addtoalist.AddToAList
+
+  type Timeline = viewmodels.timeline.Timeline
```
matthewmascord, line 105: Do we need an alias for Event too?

src/main/twirl/uk/gov/hmrc/hmrcfrontend/views/components/hmrcTimeline.scala.html
```diff
@@ -0,0 +1,33 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *@
+
+@import uk.gov.hmrc.hmrcfrontend.views.viewmodels.timeline.Timeline
+
+@this()
+
+@(params: Timeline)(implicit messages: Messages)
+@import params._
```
matthewmascord, line 22: Probably worth enforcing mandatory parameters like here: https://github.com/hmrc/play-frontend-hmrc/blob/7bc0366ad080c1c24be2d6aa328721e186e8f61d/src/main/twirl/uk/gov/hmrc/hmrcfrontend/views/components/hmrcCharacterCount.scala.html#L29

natcarey: I'm not sure what we'd test for there, `headingLevel` has a default, `events` defaults to an empty sequence.  The `Event` requires a `title`, `time` and `content` with an optional `datetime`.

src/main/scala/uk/gov/hmrc/hmrcfrontend/views/viewmodels/timeline/Event.scala
```diff
@@ -0,0 +1,30 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package uk.gov.hmrc.hmrcfrontend.views.viewmodels.timeline
+
+import play.api.libs.json.{Json, OFormat}
+
+case class Event(
+  title: String,
```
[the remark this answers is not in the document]
natcarey: I don't think there is a logical default value that we'd want in production - if we want a value to test with I'd suggest we add one when we test rather that writing one which will hit production.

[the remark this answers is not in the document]
natcarey: I don't see defaults in the `addToAList` e.g.

```
case class ListItem(
  name: String,
  changeUrl: String,
  removeUrl: String
)
```

src/main/scala/uk/gov/hmrc/hmrcfrontend/views/Aliases.scala
natcarey: Updated here https://github.com/hmrc/play-frontend-hmrc/pull/77/commits/ea82f4b7c81cccaede4d1d55f2701f17e7893337

matthewmascord (approved):

oscarduignan (approved):
looks good to me!

PR #80, opened 2021-05-27
Outputting markup as required

natcarey:
Also fixed test framework issue which incorrectly equates encoded HTML to HTML.


matthewmascord (approved):
Good to be more precise in our testing 👍

PR #93, opened 2021-06-30
PLATUI-1261: Add a RichErrorSummary

matthewmascord:


src/test/scala/uk/gov/hmrc/hmrcfrontend/views/helpers/hmrcScriptsSpec.scala
```diff
@@ -38,11 +38,12 @@ class hmrcScriptsSpec
   "hmrcScripts" should {
     "include the hmrc-frontend script tag" in {
       val hmrcScripts = app.injector.instanceOf[HmrcScripts]
+      hmrcfrontend.RoutesPrefix.setPrefix("/foo-service/hmrc-frontend")
       val scripts     = hmrcScripts().select("script")
 
       scripts                   should have size 1
       scripts.first.attr("src") should fullyMatch regex
-        """/assets/hmrc-frontend-\d+.\d+.\d+.min.js""".r
+        """/foo-service/hmrc-frontend/assets/hmrc-frontend-\d+.\d+.\d+.min.js""".r
```
oscarduignan, line 46: @matthewmascord just to check, is this change intentional or part of something else?

matthewmascord: There is a race condition around this because the reverse router is statically initialised and it was failing locally. So this forces it to have a certain value.

JoPintoPaul (dismissed):

src/main/scala/uk/gov/hmrc/hmrcfrontend/views/Implicits.scala
```diff
@@ -223,6 +223,58 @@ trait Implicits {
...
+    *
+    * For example:
+    * {{{
+    *   @govukErrorSummary(ErrorSummary().withFormTextErrors(dateInputForm, mapping = Map("date" -> "date.day")))
+    * }}}
+    */
+  implicit class RichErrorSummary[T](errorSummary: ErrorSummary)(implicit val messages: Messages) {
+    import uk.gov.hmrc.govukfrontend.views.Implicits.RichFormErrors
+
+    def withFormHtmlErrors(form: Form[T]): ErrorSummary =
```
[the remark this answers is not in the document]
matthewmascord: What about `withFormErrorsAsText` and `withFormErrorsAsHtml`?

oscarduignan (approved):

PR #98, opened 2021-07-21
PLATUI-1237: change recommended import method for components

oscarduignan:
Following discussion earlier within team we're updating our recommended import usage for components

Easiest to preview by viewing the branch itself https://github.com/hmrc/play-frontend-hmrc/tree/PLATUI-1237_change-recommended-import-method-for-components


README.md
```diff
@@ -62,22 +62,42 @@ The library comprises two packages:
...
-    follows:
-
+1. Then to use our components and helpers in your templates, you will need to import the component or helper from its
+   package `uk.gov.hmrc.(govuk|hmrc)frontend.views.html.(components|helpers)` (example import shown on the first line in the code sample below).
+
+   For most components the parameters that can be provided are encapsulated in a viewmodel, comprised of case classes within a subpackage of `uk.gov.hmrc(govuk|hmrc)frontend.views.viewmodels` (example import shown on the second line in the code sample below).
+   
+   ```scala
+   @import uk.gov.hmrc.hmrcfrontend.views.html.components.HmrcTimeline
+   @import uk.gov.hmrc.hmrcfrontend.views.viewmodels.timeline.{Timeline, Event}
```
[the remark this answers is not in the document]
oscarduignan: fixed this now!

README.md
```diff
@@ -62,22 +62,42 @@ The library comprises two packages:
...
+   ))
+   ```
+
+   > **Notice: Recommended usage pattern changed as of July 2021**
+   > 
+   > Previously instead of individual imports we suggested using wildcard imports to pull in all components and helpers (along with their implicits,) and additionally to include these imports for all templates automatically via  the `TwirlKeys.templateImports` config option in `build.sbt`.
+   > 
+   > This removed the need to add lots of individual import statements in your templates (or any at all when using `TwirlKeys.templateImports`.) The only exception being if the class being used caused an ambiguous import compilation error, which could occur because a class of the same name was provided by both wildcard imports. This error could be resolved by adding an explicit individual import for the version of the class you needed.
+   > 
+   > We no longer recommend using wildcard imports, or the TwirlKeys configuration. These introduced complexity, for example unused imports triggered error messages which needed to be suppressed, and the suppression of which might have the unintended side effect of also suppressing meaningful feedback (like deprecation warnings).
```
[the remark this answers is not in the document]
oscarduignan: @matthewmascord you're right, I've restructured it a bit now. What do you think?

natcarey (approved):

README.md
```diff
@@ -59,23 +59,58 @@ The library comprises two packages:
...
+1. Then to use our components and helpers in your templates, you will need to import the component or helper from its
+   package `uk.gov.hmrc.(govuk|hmrc)frontend.views.html.(components|helpers)`.
+
+   For most components the parameters that can be provided are encapsulated in a viewmodel, comprised of case classes
+   that live within a subpackage of `uk.gov.hmrc(govuk|hmrc)frontend.views.viewmodels` and are aliased for use
+   under `uk.gov.hmrc(govuk|hmrc)frontend.views.html.components`.
+
+   ```scala
+   /*
+   Import all components, viewmodels, and their implicits. Most succinct import method, but may require additional
```
matthewmascord, line 71: I'm not sure this is correct. The components._ would not bring in any implicit classes as far as I can see.

README.md
```diff
@@ -59,23 +59,58 @@ The library comprises two packages:
...
+         datetime = Some("2020-07-15"),
+         content = "Example content"
+       )
+     )
+   ))
+   ```
+
+   > **Notice: Recommended usage pattern changed as of July 2021**
+   >
+   > Previously instead of explicit imports within templates we suggested pulling in all components and helpers (along
```
matthewmascord, line 102: Do we need to mention what we previously recommended? Is this adding any value?

README.md
```diff
@@ -59,23 +59,58 @@ The library comprises two packages:
...
+   > with their implicits) using wildcard imports via `TwirlKeys.templateImports` in `build.sbt`.
+   >
+   > This reduces the boilerplate needed within templates. With the exception being if the class being used caused an 
+   > ambiguous import compilation error, which could occur because a class of the same name was provided by multiple 
+   > imports. This error could be resolved by adding an explicit individual import for the version of the class you 
+   > needed in the template where it was used.
+   >
+   > We no longer recommend the use of the `TwirlKeys.templateImports` configuration. This introduced some complexity, 
+   > and unused imports triggered excessive warning messages which needed to be suppressed or could cause alert fatigue.
+   > However when suppressed there was a possible undesirable side effect of also suppressing meaningful feedback like
```
matthewmascord, line 112: This is the kind of context we can add into the ADR, I don't know if it's necessary to add into the README.

README.md
matthewmascord:
```suggestion
   Import all components and viewmodels. Most succinct import method, but may require additional 
```

oscarduignan: refactored! I have included helpers as well even though they're not used to the first example so it's clearer what's needed to be able to pull in everything.

README.md
oscarduignan: trimmed down!

README.md
oscarduignan: I've kept a note here that recommendation the has changed not to avoid using the twirl keys config but I've trimmed it down. I've added the ADR in now too but seemed like a snippet here for anyone who won't read through our ADRs might help if someone is sure they saw it somewhere but then can't remember where.

README.md
```diff
@@ -59,23 +59,50 @@ The library comprises two packages:
...
+     events = Seq(
+       Event(
+         title = "Event 1",
+         time = "14 July 2020",
+         datetime = Some("2020-07-15"),
+         content = "Example content"
+       )
+     )
+   ))
+   ```
```
JoPintoPaul, line 100: As discussed, might be more useful to teams to use a more widely used component, for example `CharacterCount` from `play-frontend-hmrc`. Additionally, if using a form input component, then can demonstrate the use of the implicit `withFormField` method, as currently the import of the implicits doesn't bring anything in that actually gets used.

JoPintoPaul (approved):

PR #102, opened 2021-07-28
PLATUI-1300: Adding additionalBeforeContentBlock and  backLinkUrl to …

JoPintoPaul:
…hmrcLayout


oscarduignan (approved):

src/main/twirl/uk/gov/hmrc/hmrcfrontend/views/helpers/hmrcLayout.scala.html
```diff
@@ -26,10 +30,12 @@
         serviceUrl: Option[String] = None,
         userResearchBannerUrl: Option[String] = None,
         accessibilityStatementUrl: Option[String] = None,
+        backLinkUrl: Option[String] = None,
         displayHmrcBanner: Boolean = false,
         phaseBanner: Option[PhaseBanner] = None,
         additionalHeadBlock: Option[Html] = None,
         additionalScriptsBlock: Option[Html] = None,
+        additionalBeforeContentBlock: Option[Html] = None,
```
matthewmascord, line 38: I was wondering if we should make this an either/or - if they specify a beforeContentBlock it replaces the language toggle and back link? Alternatively could we defer adding this until we know of a use case?

matthewmascord (approved):

PR #106, opened 2021-08-09
PLATUI-1290: add HmrcPageHeadingLabel config helper

oscarduignan:
draft for some feedback

### DIFF
--- a/src/main/scala/uk/gov/hmrc/hmrcfrontend/views/config/HmrcPageHeadingLabel.scala
+++ b/src/main/scala/uk/gov/hmrc/hmrcfrontend/views/config/HmrcPageHeadingLabel.scala
@@ -1,0 +1,16 @@
+package uk.gov.hmrc.hmrcfrontend.views.config
+
+import play.twirl.api.HtmlFormat
+import uk.gov.hmrc.govukfrontend.views.viewmodels.content.HtmlContent
+import uk.gov.hmrc.govukfrontend.views.viewmodels.label.Label
+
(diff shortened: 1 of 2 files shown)


src/main/scala/uk/gov/hmrc/hmrcfrontend/views/config/HmrcPageHeadingLabel.scala
```diff
@@ -0,0 +1,16 @@
+package uk.gov.hmrc.hmrcfrontend.views.config
+
+import play.twirl.api.HtmlFormat
+import uk.gov.hmrc.govukfrontend.views.viewmodels.content.HtmlContent
+import uk.gov.hmrc.govukfrontend.views.viewmodels.label.Label
+
+object HmrcPageHeadingLabel {
```
matthewmascord, line 7: I guess my only question is whether it needs the Hmrc prefix. We didn't for StandardPhaseBanner but then perhaps we should have? 🤷

oscarduignan: yeah, I wasn't sure either way, I defaulted to using it because it seems like the recommendation is slightly different to the govuk one https://design-system.service.gov.uk/styles/typography/captions-inside/index.html

oscarduignan:
closing this, superseded by another branch https://github.com/hmrc/play-frontend-hmrc/pull/111

PR #109, opened 2021-08-11
Removed the deprecated static helpers.

natcarey:


matthewmascord (commented):
One of the ACs was for the version number to be increased to 1.0.0.

I think the play-frontend-govuk changes need to be made first, otherwise we'll end up with only half of the static helpers being removed.

Could we squash commits please?

src/it/scala/uk/gov/hmrc/hmrcfrontend/support/TemplateIntegrationSpec.scala
```diff
@@ -39,6 +45,8 @@ abstract class TemplateIntegrationSpec[T: OWrites: Arbitrary](hmrcComponentName:
   override def overrideParameters(p: Test.Parameters): Test.Parameters =
     p.withMinSuccessfulTests(20)
 
+  val component = app.injector.instanceOf[C]
```
oscarduignan, line 48: @natcarey with this refactoring could we now move the render method to the abstract class too? Seems like it's now basically the same code except for the type of the render param but we have that in T

natcarey: Ooh, that would be lovely.  I'll try that. :)

natcarey:
Yeah, I'm holding off the 1.0.0 until we've merged `play-frontend-govuk` in - that's particularly important if we haven't ruled out a name change.

docs/maintainers/adr/0013-removing-deprecated-helpers.md
```diff
@@ -0,0 +1,37 @@
+# play-frontend-hmrc mirrors hmrc-frontend
```
matthewmascord, line 1: Title needs updating.

docs/maintainers/adr/0013-removing-deprecated-helpers.md
```diff
@@ -0,0 +1,37 @@
...
+
+* Status: accepted
+* Date: 2020-06-24
```
matthewmascord, line 4: Date needs updating.

natcarey: The date is based on the comment in Jira documenting the decision when we made it.

docs/maintainers/adr/0013-removing-deprecated-helpers.md
natcarey: Updated.

matthewmascord:
Superceded by https://github.com/hmrc/play-frontend-hmrc/pull/112

PR #113, opened 2021-08-19
PLATUI-1320: Inline play-frontend-govuk into this library

matthewmascord:
Inline the play-frontend-govuk library as per recent [ADR](https://github.com/hmrc/play-frontend-hmrc/blob/master/docs/maintainers/adr/0011-inline-play-frontend-govuk.md).

This PR focusses on merging the two repositories, resolving any clashes (e.g. fixture generation), and updating the documentation.

Subsequent PRs will focus on removing further duplication. For example, unit test and implicit support.

This is not a breaking change for consumers. However, consumers will need to ensure they remove any direct references to play-frontend-govuk in their `project/AppDependencies.scala` file. Not doing so could result in outdated GOV.UK components being loaded.


oscarduignan:
It doesn't want me viewing the changes on github! Will try checking it out today and browsing around

matthewmascord:
This is the most relevant [commit](https://github.com/hmrc/play-frontend-hmrc/commit/b8847b697e7628b993ca21b44633b88703f25beb) @oscarduignan  - this is the commit that resolves the clashes between the two repositories.

oscarduignan (approved):
I've looked through this a few times now and it looks good to me, but be good to get more eyes on it because it's big

docs/maintainers/overview.md
```diff
@@ -74,65 +89,94 @@ the Twirl template output and the `new` file was the Nunjucks template output (e
...
 
+### Running all tests for all supported Scala and Play versions
+
+Prior to merging to trunk, it is a good idea to run all the tests against all supported versions of Scala and Play.
+These checks will also be performed automatically as part of the build pipeline before publishing.
+
+To achieve this, run:
+
+```bash
+PLAY_VERSION=2.6 sbt clean +test +it:test && \
```
JoPintoPaul, line 100: This reference to Play 2.6 should be removed.

JoPintoPaul (approved):

PR #114, opened 2021-08-25
PLATUI-1035_reduce_nesting

matthewmascord:


src/main/scala/uk/gov/hmrc/govukfrontend/views/implicits/RichStringSupport.scala
```diff
@@ -16,10 +16,47 @@
...
+import uk.gov.hmrc.govukfrontend.views.viewmodels.hint.Hint
+import uk.gov.hmrc.govukfrontend.views.viewmodels.label.Label
+import uk.gov.hmrc.govukfrontend.views.viewmodels.summarylist.Key
+import uk.gov.hmrc.govukfrontend.views.viewmodels.table.HeadCell
+import uk.gov.hmrc.govukfrontend.views.viewmodels.tag.Tag
 
+trait RichStringSupport {
   implicit class RichString(s: String) {
 
+    def toHint(args: Any*)(implicit messages: Messages): Hint = Hint(content = Text(messages(s, args)))
```
matthewmascord, line 31: This demonstrates adding an extension method to convert rather than relying on implicit conversions - less magical. Also allows for addition of message arguments.

src/main/scala/uk/gov/hmrc/govukfrontend/views/implicits/RichTextareaSupport.scala
```diff
@@ -54,6 +56,22 @@ trait RichTextareaSupport {
         .withValue(field)
         .withHtmlErrorMessage(field)
 
+    // FIXME: PoC only - needs tests and functionality to not-override any existing label
+    def withHeadingAndSectionCaption(heading: Content, sectionCaption: Content): Textarea =
```
matthewmascord, line 60: Simplifies boilerplate needed to follow the 1-question per page pattern when using a section caption. Harder to do with implicit conversions due to the need for two parameters.

src/main/scala/uk/gov/hmrc/hmrcfrontend/views/config/HmrcYesNoRadioItems.scala
```diff
@@ -0,0 +1,29 @@
...
+ */
+
+package uk.gov.hmrc.hmrcfrontend.views.config
+
+import play.api.i18n.Messages
+import uk.gov.hmrc.govukfrontend.views.viewmodels.content.Text
+import uk.gov.hmrc.govukfrontend.views.viewmodels.radios.RadioItem
+
+object HmrcYesNoRadioItems {
+  def apply()(implicit messages: Messages): Seq[RadioItem] =
```
matthewmascord, line 24: This just encapsulates the standard Yes/No radios seen in multiple places across the platform.

oscarduignan:
@matthewmascord these all make sense to me, though I don't know if I have enough usage experience with scala implicits and consuming the play-frontend libraries to catch any subtleties about adding these. If we have it turned on by default, could it conflict with any other implicit conversion helpers teams might have created? I wonder if we could find an example and test it out to see how it behaves - or introduce this first by services having to opt-in

JoPintoPaul (commented):
Interesting suggestions. My main thoughts are the following: 
1. Specific implementations - I see that the majority of these implicit conversions assume that the Content type will be `Text` rather than `Html`. I would be interested to know whether teams do have requirements about passing through HTML, as per the `&nbsp` issue seen with https://github.com/hmrc/play-frontend-hmrc/blob/master/src/main/scala/uk/gov/hmrc/govukfrontend/views/implicits/ImplicitsSupport.scala#L31
2. General design decision - assuming strings lifted by implicit conversions are message keys - I'd be interested to discuss this design decision.
3. Implicit conversions in general - looking this, I'd be keen to have a think about whether the code becomes a bit "magic" by these helpers - I would wonder whether developers looking at existing code would struggle to understand the conversions because it can be very difficult to  debug which implicit conversion is called into scope in code.

PR #124, opened 2021-09-09
PLATUI-1336: Consolidating duplicated code following merge of play-fr…

JoPintoPaul:
…ontend-hmrc and play-frontend-govuk


src/main/scala/uk/gov/hmrc/hmrcfrontend/views/config/HmrcPageHeading.scala
```diff
@@ -17,7 +17,7 @@
 package uk.gov.hmrc.hmrcfrontend.views.config
 
 import play.twirl.api.{Html, HtmlFormat}
-import uk.gov.hmrc.govukfrontend.views.Utils.toClasses
+import uk.gov.hmrc.supportfrontend.views.Utils.toClasses
```
matthewmascord, line 20: Presumably slight risk teams could be using Utils directly?

src/main/scala/uk/gov/hmrc/supportfrontend/views/Utils.scala
```diff
@@ -46,20 +48,23 @@ trait Utils {
...
-  }
-
-  private[views] def isNonEmptyOptionString(value: Option[String]) = value match {
+  def isNonEmptyOptionString(value: Option[String]): Boolean = value match {
     case Some(NonEmptyString(_)) => true
     case _                       => false
   }
 
-  private[views] def calculateAssetPath(path: Option[String], file: String): String =
+  def calculateAssetPath(path: Option[String], file: String): String =
```
matthewmascord, line 56: Is there a reason to make this public?

src/main/scala/uk/gov/hmrc/hmrcfrontend/views/config/HmrcPageHeading.scala
JoPintoPaul: Hmm, good point. We could make this non-breaking by having this code in a trait in a trait and preserving the original companion objects. Shall I make this change?

src/main/scala/uk/gov/hmrc/supportfrontend/views/Utils.scala
JoPintoPaul: This will no longer work because the package structure has changed. Previously this method was private to the package `uk.gov.hmrc.govukfrontend.views`. However, because this method is now in the package `uk.gov.hmrc.supportfrontend`, it cannot be made private to a different package of `uk.gov.hmrc.govukfrontend` (see this post on SO: https://stackoverflow.com/questions/9032631/package-private-modifier-in-scala-2-8

src/main/scala/uk/gov/hmrc/hmrcfrontend/views/config/HmrcPageHeading.scala
matthewmascord: With a deprecation notice?

src/main/scala/uk/gov/hmrc/supportfrontend/views/Utils.scala
matthewmascord: calculateAssetPath is really private to GovukTemplate and govukfrontend - it's not used by the hmrc components at all. Can it privately live in govukfrontend.components?

matthewmascord (approved):

PR #127, opened 2021-09-10
PLATUI-1336: Consolidating Assets controller

JoPintoPaul:


src/main/twirl/uk/gov/hmrc/govukfrontend/views/components/GovukHeader.scala.html
```diff
@@ -54,7 +54,7 @@
           </svg>
           <!--<![endif]-->
           <!--[if IE 8]>
-          <img src="@calculateAssetPath(assetsPath, "images/govuk-logotype-crown.png")" class="govuk-header__logotype-crown-fallback-image" width="36" height="32">
+          <img src="@routes.Assets.at("images/govuk-logotype-crown.png")" class="govuk-header__logotype-crown-fallback-image" width="36" height="32">
```
matthewmascord, line 57: Ah, I think I had intended that instead of using the images in the resources directory we would remove those images and use the ones in the hmrcfrontend webjar instead (via the reverse router)

JoPintoPaul: @matthewmascord Worry not, this is still a WIP! I am making that change now to delete the images in the `/assets` folder, and use the ones from `govuk-frontend` instead.

PR #141, opened 2021-10-12
Update LibDependencies.scala

oscarduignan:
checking if we can remove two dependencies

- joda-convert I couldn't see usage, and I didn't seem to get any errors removing it, so wanted to check if this was safe to remove because this was getting evicted by services version
- scalaj-http we are using to generate fixtures and have in the [build.sbt](https://github.com/hmrc/play-frontend-hmrc/blob/main/project/build.sbt#L2) in our project folder but I don't think we have it in src/main anywhere so if we remove this it avoids some possible evictions when integrating with services
- play ws is only used by our integration tests so I scoped it to Test in case that reduces what's pulled in, but it might just be a part of our dependency on play I guess


JoPintoPaul (approved):

src/test/resources/fixtures/hmrc-frontend/additional-fixtures/new-tab-link-no-language/output.txt
```diff
@@ -1 +1 @@
-<a class="govuk-link " target="_blank" rel="noopener noreferrer" href="https://www.gov.uk/working-tax-credit/eligibility">find out if you qualify for tax credits</a>
+<a class="govuk-link " target="_blank" rel="noopener noreferrer" href="https://www.gov.uk/working-tax-credit/eligibility">find out if you qualify for tax credits </a>
```
oscarduignan, line 1: when I was running my test of generate fixtures for scalaj this popped up as a small change for the fixtures so I have included it

PR #144, opened 2021-11-03
PLATUI-1419: updated the account menu to match latest version from PTA

oscarduignan:
Co-authored-by: Kyle Bowden <1393788+kyle-bowden@users.noreply.github.com>


build.sbt
```diff
@@ -11,7 +11,7 @@ lazy val root = Project(libName, file("."))
   .configs(IntegrationTest)
   .settings(
     name := libName,
-    majorVersion := 1,
+    majorVersion := 2,
```
JoPintoPaul, line 14: Ooh, major version change. SO, I know this is a breaking change for the PTA teams, but I think it would be worth discussing whether we think this qualifies as a major version. Last time we did a major version it was a LOT of architecture changes IIRC, and we don't always major version breaking changes.

src/main/twirl/uk/gov/hmrc/hmrcfrontend/views/components/HmrcAccountMenu.scala.html
```diff
@@ -69,50 +69,12 @@
     </li>
 
     <li>
-      <a href="#subnav-your-account" id="account-menu__main-2" class="hmrc-account-menu__link hmrc-account-icon hmrc-account-menu__link--more" aria-haspopup="true" aria-owns="subnav-your-account">
+      <a href="@yourProfile.href" class="hmrc-account-menu__link@if(yourProfile.active){ hmrc-account-menu__link--active}else{}">
```
JoPintoPaul, line 72: Do we need the `else` block here?

oscarduignan: ah maybe not, we just copy pasted from the others I think, we can try it out

build.sbt
oscarduignan: Hmm, I think going from v0 to v1 is quite special because it's somewhat saying "now we'll really try to avoid breaking changes / api is pretty stable" and until v1 it feels like maybe you can introduce breaking changes in minor versions, but past v1 I wonder if we should switch to bump it whenever there's a breaking change no matter how big or small (though maybe we haven't clearly defined what our API surface is)

src/main/twirl/uk/gov/hmrc/hmrcfrontend/views/components/HmrcAccountMenu.scala.html
oscarduignan: that worked, we've remove it from this and the other places 👍

build.sbt
oscarduignan: we've updated the readme to explain that the account menu is currently experimental and being developed outside this repo by PTA so subject to breaking API changes 👍

JoPintoPaul (approved):

PR #149, opened 2021-11-24
Automatically populate links to services that are consistent across all usages of the pta account menu

kyle-bowden:
Part of the acceptance criteria is to get PTA to review this ticket along with us.

I will tag in @Muhammad-Ali-Ahmad  to have a look at the code changes.


src/main/scala/uk/gov/hmrc/hmrcfrontend/config/AccountMenuConfig.scala
```diff
@@ -0,0 +1,43 @@
...
+  private val platformHost: Option[String] =
+    configuration.getOptional[String]("platform.frontend.host")
+
+  private val ptaAccountMenuHost: Option[String] =
+    configuration.getOptional[String]("pta-account-menu.host")
+
+  private def baseUrl: String =
+    platformHost
+      .orElse(ptaAccountMenuHost)
+      .getOrElse("")
```
oscarduignan, line 33: @JoPintoPaul / @kyle-bowden I don't know if this is going against the convention of the platform, but should we move the default value into the config instead? And maybe have separate hosts for the different services rather than a shared one, otherwise if you are running the services locally on different ports then some / all won't work

```
pta-account-menu.pertax-frontend.host = ${platform.frontend.host:-localhost:0000}
```

default value there is made up but just to show what I mean, I'm not sure if pertax-frontend is the right service name either here

kyle-bowden: Thank you for the suggestion, the code has now been updated to reflect the change.

src/main/resources/reference.conf
```diff
@@ -37,4 +37,15 @@ hmrc-timeout-dialog {
...
 
+pta-account-menu {
+  pertax-frontend {
+    host = "http://localhost:9232"
+    host = ${?platform.frontend.host}
+  }
+  accountHomePath = "/personal-account"
+  messagesPath = "/personal-account/messages"
+  checkProgressPath = "/track"
+  yourProfilePath = "/personal-account/your-profile"
```
Muhammad-Ali-Ahmad, line 48: Is business-account URL part of another ticket?

kyle-bowden: Yes, its part of [PLATUI-1468](https://jira.tools.tax.service.gov.uk/browse/PLATUI-1468) which we will be looking at next.

src/main/scala/uk/gov/hmrc/hmrcfrontend/config/AccountMenuConfig.scala
```diff
@@ -0,0 +1,40 @@
...
+    configuration.getOptional[String]("pta-account-menu.pertax-frontend.host")
+
+  private def baseUrl: String = perTaxFrontendHost.getOrElse("")
+
+  val accountHomeHref: String   =
+    s"$baseUrl${configuration.get[String]("pta-account-menu.accountHomePath")}"
+  val messagesHref: String      =
+    s"$baseUrl${configuration.get[String]("pta-account-menu.messagesPath")}"
+  val checkProgressHref: String =
+    s"$baseUrl${configuration.get[String]("pta-account-menu.checkProgressPath")}"
```
oscarduignan, line 34: @kyle-bowden I think this should have a different base url, it's only locally that they're not the same, but this might be quite an important component to be able to have working locally. This is tracking-frontend I think, @Muhammad-Ali-Ahmad might be able to confirm

src/main/twirl/uk/gov/hmrc/hmrcfrontend/views/helpers/HmrcAccountMenuHelper.scala.html
```diff
@@ -0,0 +1,33 @@
...
+@import uk.gov.hmrc.hmrcfrontend.views.viewmodels.accountmenu._
+@import uk.gov.hmrc.hmrcfrontend.views.viewmodels.language.{Cy, En}
+
+@this(hmrcAccountMenu: HmrcAccountMenu, accountMenuConfig: AccountMenuConfig)
+
+@(signOutHref: String)(implicit messages: Messages, request: RequestHeader)
+
+@hmrcAccountMenu(AccountMenu(
+  accountHome = AccountHome(href = accountMenuConfig.accountHomeHref),
+  messages = AccountMessages(href = accountMenuConfig.messagesHref),
```
oscarduignan, line 28: @kyle-bowden I'm sorry I've just realised I've lost track of one question I had that we didn't answer. @Muhammad-Ali-Ahmad the account messages can have a count on it https://github.com/hmrc/play-frontend-hmrc/blob/main/src/main/scala/uk/gov/hmrc/hmrcfrontend/views/viewmodels/accountmenu/AccountMessages.scala#L24 but we don't know how to populate it, can we pass through a parameter for it from this helper for now?

oscarduignan: @kyle-bowden I think another thing that's just clicked here for me is that we also need to be able to toggle the "active" status of the links - could we possibly do this automatically by checking the current request path, @Muhammad-Ali-Ahmad , wiring that in automatically might need to be done in a future change with the message count, and for now just allow it to be set

src/main/scala/uk/gov/hmrc/hmrcfrontend/config/AccountMenuConfig.scala
kyle-bowden: I have introduced the `pertax-frontend.host` configuration, so that this can be set explicitly if needed.

src/main/twirl/uk/gov/hmrc/hmrcfrontend/views/helpers/HmrcAccountMenuHelper.scala.html
kyle-bowden: Both of these suggestions are now available as features to the AccountMenu.

src/main/scala/uk/gov/hmrc/hmrcfrontend/config/AccountMenuConfig.scala
oscarduignan: Ah, @kyle-bowden sorry, pertax-frontend makes sense to me, but not all of these urls are provided by pertax-frontend, this one (check progress) is from https://github.com/hmrc/tracking-frontend I think (https://github.com/hmrc/tracking-frontend/blob/main/conf/prod.routes#L2)

kyle-bowden: So i have added config to override the hosts for individual links, by default it will use the `pta-account-menu.pertax-frontend.host` but they can be overridden in each service link configuration as in the example below.

`pta-account-menu {
  accountHome = {
    host = "http://localhost:12345"
  }

  messages = {
    host = "http://localhost:12346"
  }

  checkProgress = {
    host = "http://localhost:12347"
  }

  yourProfile = {
    host = "http://localhost:12348"
  }
}`

src/main/resources/reference.conf
```diff
@@ -37,4 +37,31 @@ hmrc-timeout-dialog {
...
+  checkProgress = {
+    host = ${?pta-account-menu.pertax-frontend.host}
+    path = "/track"
+  }
+
+  yourProfile = {
+    host = ${?pta-account-menu.pertax-frontend.host}
+    path = "/personal-account/your-profile"
+  }
+}
```
oscarduignan, line 65: What about this? I've taken the tracking-frontend port from [pertax-frontend](https://github.com/hmrc/pertax-frontend/blob/main/conf/application.conf#L240-L242) and I changed the menu keys to be kebab-case delimited rather than camelCase to make it consistent within our config block, could go either way - not sure what the convention is here but figure we should be consistent
```suggestion
pta-account-menu {
  pertax-frontend {
    host = "http://localhost:9232"
    host = ${?platform.frontend.host}
  }

  tracking-frontend {
    host = "http://localhost:9100"
    host = ${?platform.frontend.host}
  }

  account-home = {
    host = ${pta-account-menu.pertax-frontend.host}
    path = "/personal-account"
  }

  messages = {
    host = ${pta-account-menu.pertax-frontend.host}
    path = "/personal-account/messages"
  }

  check-progress = {
    host = ${pta-account-menu.tracking-frontend.host}
    path = "/track"
  }

  your-profile = {
    host = ${pta-account-menu.pertax-frontend.host}
    path = "/personal-account/your-profile"
  }
}
```

oscarduignan: if you do this, then you can also change the getOptionals below and just have get string, because the fallback stuff happens in the reference conf, not duplicated in the code

src/main/scala/uk/gov/hmrc/hmrcfrontend/config/AccountMenuConfig.scala
```diff
@@ -0,0 +1,48 @@
...
+
+  private val perTaxFrontendHost: Option[String] =
+    configuration.getOptional[String]("pta-account-menu.pertax-frontend.host")
+
+  private def baseUrl: String = perTaxFrontendHost.getOrElse("")
+
+  val accountHomeHref: String = {
+    val accountMenuHost = configuration.getOptional[String]("pta-account-menu.accountHome.host").getOrElse(baseUrl)
+    s"$accountMenuHost${configuration.get[String]("pta-account-menu.accountHome.path")}"
+  }
```
oscarduignan, line 32:
```suggestion
  val accountHomeHref: String = 
    s"$configuration.get[String]("pta-account-menu.account-home.host")${configuration.get[String]("pta-account-menu.account-home.path")}"
```
example of how it could simplify this, so remove the default values in code and just use them in the config

src/main/resources/reference.conf
```diff
@@ -37,4 +37,36 @@ hmrc-timeout-dialog {
...
+  }
+
+  check-progress = {
+    host = ${pta-account-menu.tracking-frontend.host}
+    path = ${pta-account-menu.check-progress.host}"/track"
+  }
+
+  your-profile = {
+    host = ${pta-account-menu.pertax-frontend.host}
+    path = ${pta-account-menu.your-profile.host}"/personal-account/your-profile"
```
oscarduignan, line 68: @kyle-bowden tiny thing, but could we rename these from path to href, matches the config vars then now that it has the host in it too

src/main/twirl/uk/gov/hmrc/hmrcfrontend/views/components/HmrcAccountMenu.scala.html
```diff
@@ -14,7 +14,9 @@
  * limitations under the License.
  *@
 
-@this(hmrcNotificationBadge: HmrcNotificationBadge)
+@import uk.gov.hmrc.hmrcfrontend.config.AccountMenuConfig
+
+@this(hmrcNotificationBadge: HmrcNotificationBadge)(implicit accountMenuConfig: AccountMenuConfig)
```
oscarduignan, line 19: @kyle-bowden I think we can remove this implicit now that we changed to a helper

src/main/resources/reference.conf
kyle-bowden: Sure, i will make that change.

oscarduignan (approved):

PR #150, opened 2021-11-30
Plat UI 1470 internal gtm config

JoPintoPaul:


kyle-bowden (approved):

src/main/twirl/uk/gov/hmrc/hmrcfrontend/views/helpers/HmrcInternalHead.scala.html
```diff
@@ -0,0 +1,31 @@
...
+@this(hmrcInternalGtmScript: HmrcInternalGtmScript, assetsConfig: AssetsConfig)
+
+@(headBlock: Option[Html] = None, nonce: Option[String] = None)(implicit request: RequestHeader, messages: Messages)
+@hmrcInternalGtmScript(nonce)
+<!--[if lte IE 8]>
+<script src="@assetsConfig.html5ShivJsUrl"></script>
+<link href="@assetsConfig.hmrcFrontendIe8CssUrl" media="all" rel="stylesheet" type="text/css" />
+<![endif]-->
+<!--[if gt IE 8]><!-->
+<link href="@assetsConfig.hmrcFrontendCssUrl" media="all" rel="stylesheet" type="text/css"@nonce.map {n=> nonce="@n"} />
```
oscarduignan, line 29: @JoPintoPaul looks good to me, and if no csp changes are needed, only thought I have is should we also include in the guidance a note about not using this alongside HmrcHead, that it's an alternative for internal services / maybe that if you are using HmrcLayout to use HmrcInternalGtmScript directly or make sure to swap HmrcHead for HmrcInternalHead. Not sure it would break anything but they would be including stylesheets twice

oscarduignan (approved):

PR #156, opened 2022-01-28
PLATUI-1533: Upgraded versions of govuk-frontend and hmrc-frontend, a…

JoPintoPaul:
…nd removed references to Play 2.6 and Play 2.7


src/main/twirl/uk/gov/hmrc/govukfrontend/views/components/GovukSummaryList.scala.html
```diff
@@ -20,7 +20,7 @@
 @import params._
 <dl class="@toClasses("govuk-summary-list", classes)"@toAttributes(attributes)>
   @for(row <- rows) {
-    <div class="@toClasses("govuk-summary-list__row", row.classes)">
+    @if(anyRowHasActions && row.actions.isEmpty) {<div class="@toClasses("govuk-summary-list__row", "govuk-summary-list__row--no-actions", row.classes)">} else {<div class="@toClasses("govuk-summary-list__row", row.classes)">}
```
oscarduignan, line 23: Small optional one from refactoring from me, noticed that some other places we are using a string.when(predicate) implicit which looks like it's [provided by twirl](https://github.com/playframework/twirl/blob/main/api/shared/src/main/scala/play/twirl/api/TwirlHelperImports.scala#L26-L34) to default to an empty string, which toClasses() will exclude from output
```suggestion
    <div class="@toClasses("govuk-summary-list__row", "govuk-summary-list__row--no-actions".when(anyRowHasActions && row.actions.isEmpty), row.classes)">
```

oscarduignan (approved):

kyle-bowden (approved):

oscarduignan (approved):

PR #157, opened 2022-01-28
Added HmrcPageHeadingHelper component which allows the language parameter to be passed in implicitly

kyle-bowden:


src/main/twirl/uk/gov/hmrc/hmrcfrontend/views/helpers/HmrcPageHeadingHelper.scala.html
```diff
@@ -0,0 +1,29 @@
...
+
+@this(hmrcNewTabLink: HmrcPageHeading)
+
+
+@(pageHeadingHelper: PageHeadingHelper)(implicit messages: Messages)
+@hmrcNewTabLink(PageHeading(
+  text = pageHeadingHelper.text,
+  section = pageHeadingHelper.section,
+  language = Some(messages.lang.language)
+))
```
oscarduignan, line 29: @kyle-bowden @JoPintoPaul it feels like we are adding a lot of extra stuff to the API for developers for this, and I wonder if we should consider an implicit or a helper for constructing just the viewmodel, so for example something like one of these

```scala
hmrcPageHeading(LocalisedPageHeading(text="page title"))
hmrcPageHeading(PageHeading(text="page title").withLanguage)
```

kyle-bowden: It makes sense to me to actually go this way instead of adding an additional helper component.

I like the bellow approach 
`hmrcPageHeading(PageHeading(text="page title").withLanguage)`

oscarduignan: I agree, I think with an implicit like that you get some discoverability of the API via IDE completions, where as you have to know about the helpers and remember the when you need them. I guess also it doesn't actually need to be an implicit, pretty small change then too, don't really even need an implicit (like below), but could do it like that still for consistency

```
case class PageHeading(title: String, language: String) {
  def withLanguage(implicit lang: Lang): PageHeading = this.copy(
    language = lang.code match {
      case "cy" => "cy"
      case _    => "en"
    }
  )
}
```

oscarduignan: (that code is total psuedocode, can't remember how to get language code out and I know we have En / Cy classes not strings)

JoPintoPaul (approved):

PR #161, opened 2022-02-28
Update integration guidance for timeout dialog

oscarduignan:
Couldn't find any definitive / linkable guidance around using the sign out routes provided by bas-gateway. So have not added any information about general sign out behaviour on platform. However I've tried to adjust the language so it's clearer we're not providing any suitable default value from the library itself. But that there might be a value in use already by the service that a developer can probably reuse.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,7 @@ For compatibility information see `govukFrontendVersion` and `hmrcFrontendVersio
 
 ### Changed
 
-- Tried to improve integration guidance for hmrc timeout dialog in README around use of signOutUrl
+- Updated integration guidance for hmrc timeout dialog in README around use of signOutUrl
 
 ### Compatible with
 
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -8,6 +8,17 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 For compatibility information see `govukFrontendVersion` and `hmrcFrontendVersion` in
 [LibDependencies](project/LibDependencies.scala)
 
+## [3.5.0] - 2022-02-28
+
+### Changed
+
+- Tried to improve integration guidance for hmrc timeout dialog in README around use of signOutUrl
```
JoPintoPaul, line 15:
```suggestion
- Updated integration guidance for hmrc timeout dialog in README around use of signOutUrl
```

You've more than tried! :D

JoPintoPaul (approved):

PR #166, opened 2022-03-31
PLATUI-1616: implement the hmrc list-with-actions component using a govuk summary-list

kyle-bowden:


CHANGELOG.md
```diff
@@ -8,6 +8,18 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 For compatibility information see `govukFrontendVersion` and `hmrcFrontendVersion` in
 [LibDependencies](project/LibDependencies.scala)
 
+## [4.0.0] - 2022-03-31
+
+### Changed
+
+- Update the list-with-actions component and added the ability to add custom classes to the row element
+- Updated hmrc-frontend to 5.0.1
```
JoPintoPaul, line 16: Looking at the `LibDependencies`, I think this is actually `5.0.2`, not `5.0.1`.

src/main/scala/uk/gov/hmrc/hmrcfrontend/views/viewmodels/listwithactions/ListWithActions.scala
```diff
@@ -24,7 +24,9 @@ case class ListWithActions(
   items: Seq[ListWithActionsItem] = Nil,
   classes: String = "",
   attributes: Map[String, String] = Map.empty
-)
+) {
+  val anyRowHasActions: Boolean = items.exists(_.actions.nonEmpty)
```
JoPintoPaul, line 28: Might be worth a chat about this. Historically, we have tended to put this sort of logic into the Twirl template, so that logic isn't split between viewmodel and view, but in one or the other. I would prefer if we could stick to that convention, as the rest of the display logic is in the `HmrcListWithActions.scala.html` template.

src/main/scala/uk/gov/hmrc/hmrcfrontend/views/viewmodels/listwithactions/ListWithActionsItem.scala
```diff
@@ -21,7 +21,8 @@ import uk.gov.hmrc.govukfrontend.views.viewmodels.content.{Content, Empty}
 
 final case class ListWithActionsItem(
   name: Content = Empty,
-  actions: Seq[ListWithActionsAction] = Nil
+  actions: Seq[ListWithActionsAction] = Nil,
+  classes: Option[String]
```
JoPintoPaul, line 25: Does this need a default value?

src/main/twirl/uk/gov/hmrc/hmrcfrontend/views/components/HmrcAddToAList.scala.html
```diff
@@ -76,7 +76,8 @@ <h1 class="govuk-heading-xl">
                 visuallyHiddenText = Some(if (language == Cy) s"Dileu’r ${item.name} o’r rhestr" else s"Remove ${item.name} from the list"),
                 href= item.removeUrl
             )
-          ))),
+          ),
+          classes = None)),
```
JoPintoPaul, line 80: As above, I think the default should be in the viewmodel not the view to follow library convention.

src/main/twirl/uk/gov/hmrc/hmrcfrontend/views/components/HmrcListWithActions.scala.html
```diff
@@ -17,25 +17,40 @@
...
 @import uk.gov.hmrc.hmrcfrontend.views.viewmodels.listwithactions.ListWithActionsAction
 @import uk.gov.hmrc.govukfrontend.views.viewmodels.content.NonEmptyHtml
+@import uk.gov.hmrc.hmrcfrontend.views.viewmodels.listwithactions.ListWithActionsItem
 
 @this()
 
 @(params: ListWithActions)
 @import params._
+
+@maybeNoActions(item: ListWithActionsItem) = @{
```
JoPintoPaul, line 27: Naming: I might prefer to see this called something like `noActionsClasses`, to be a bit more specific.

src/test/resources/fixtures/hmrc-frontend/test-fixtures/list-with-actions-empty-list/output.txt
```diff
@@ -0,0 +1,2 @@
+
```
JoPintoPaul, line 1: Just to confirm - when there is an empty list of Actions, should output an empty string?

kyle-bowden: Yes, we have not made any changes to this behavior, so as far as I am aware this is the way it should be working.

JoPintoPaul (approved):

PR #168, opened 2022-04-06
PLATUI-1681: HmrcPageHeading is missing the default govuk-caption-xl style

kyle-bowden:


oscarduignan:
@kyle-bowden missing new entry in changelog

src/main/scala/uk/gov/hmrc/hmrcfrontend/views/config/HmrcPageHeading.scala
```diff
@@ -67,12 +86,12 @@ object HmrcPageHeadingLegend extends HmrcPageHeading {
               """<h1 class="govuk-fieldset__heading hmrc-page-heading govuk-!-margin-top-0 govuk-!-margin-bottom-0">"""
             ),
             content.asHtml,
-            pageHeadingCaption(caption),
+            pageHeadingCaption(caption, makeCaptionSizeClass(classes, HmrcLegendSize)),
```
oscarduignan, line 89: @kyle-bowden what do you think about renaming stuff a bit:
```
govukCaptionClassScaledToMatchHeadingSize(
  headingClasses=classes,
  regexToExtractSizeFromHeadingClasses=GovukFieldsetLegendSize
)
```

src/test/scala/uk/gov/hmrc/hmrcfrontend/views/config/HmrcPageHeadingSpec.scala
```diff
@@ -31,11 +32,11 @@ class HmrcPageHeadingSpec extends AnyWordSpec with Matchers {
         isPageHeading = true,
         content = HtmlContent(
           """What is your name? """
-            + """<span class="hmrc-caption">"""
+            + """<span class="hmrc-caption govuk-caption-xl">"""
             + """Personal details"""
             + """</span>"""
         ),
-        classes = "govuk-label--xl hmrc-page-heading govuk-!-margin-top-0 govuk-!-margin-bottom-2"
+        classes = "hmrc-page-heading govuk-!-margin-top-0 govuk-!-margin-bottom-2 govuk-label--xl"
```
oscarduignan, line 39: hopefully this reordering doesn't change the styles that get applied, I think it should be ok - I'm probably misremembering how css stuff applies and it doesn't matter

src/main/scala/uk/gov/hmrc/hmrcfrontend/views/config/HmrcPageHeading.scala
kyle-bowden: I like this, I was unsure of how to name it and how verbose I should get.

src/test/scala/uk/gov/hmrc/hmrcfrontend/views/config/HmrcPageHeadingSpec.scala
kyle-bowden: I will do a manual test by using the HmrcPageHeaderLegend & HmrcPageHeaderLabel in contact frontend for this.

kyle-bowden: Seems to be fine, the only thing I can spot is if we have a class like "govuk-fieldset__legend--m govuk-fieldset__legend--xx" the last CSS style will match the regex but xx is not a valid size?

image

oscarduignan: if we can do that easily then that sounds like a useful extra bit of protection, I guess someone might create their own custom modifier maybe - but happy either way because I think that's something services can fix and verify themselves most likely

oscarduignan (approved):

PR #170, opened 2022-04-12
PLATUI-1690 remove trailing space from HmrcNewTabLink/HmrcNewTabLinkHelper

ellamdav:
I was thinking of extracting the newTabLink into a Scala helper to make it less ugly and prone to regression, but there's no existing pattern for doing that in this repo, and these components pretty much mirror those in [hmrc-frontend](https://github.com/hmrc/hmrc-frontend/blob/main/src/components/new-tab-link/template.njk), so it's probably better to leave it as-is.

I've added a test to ensure the HmrcNewTabLinkHelper component has no trailing space, to guard against any future regression.

### DIFF
--- a/src/it/scala/uk/gov/hmrc/hmrcfrontend/views/components/HmrcAddToAListIntegrationSpec.scala
+++ b/src/it/scala/uk/gov/hmrc/hmrcfrontend/views/components/HmrcAddToAListIntegrationSpec.scala
@@ -11,7 +11,7 @@ import uk.gov.hmrc.support.TemplateIntegrationBaseSpec
 
 import scala.util.Try
 
-object HmrcAddToAListSpec
+object HmrcAddToAListIntegrationSpec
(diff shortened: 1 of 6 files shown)


src/it/scala/uk/gov/hmrc/hmrcfrontend/views/components/HmrcAddToAListIntegrationSpec.scala
```diff
@@ -11,7 +11,7 @@ import uk.gov.hmrc.support.TemplateIntegrationBaseSpec
 
 import scala.util.Try
 
-object HmrcAddToAListSpec
```
ellamdav, line 14: renamed these to match the prevailing convention, and because Intellij complained of name collisions

kyle-bowden (dismissed):

kyle-bowden (approved):

PR #188, opened 2022-09-14
PLAT-1959: Review test coverage in play-frontend-hmrc

kyle-bowden:


src/main/twirl/uk/gov/hmrc/hmrcfrontend/views/components/HmrcReportTechnicalIssue.scala.html
```diff
@@ -22,7 +22,7 @@
   <a lang="@params.language.code"
     hreflang="@params.language.code"
     class="@toClasses("govuk-link hmrc-report-technical-issue ", classes.getOrElse(""))"
-    @if(params.referrerUrl.nonEmpty) { rel="noreferrer noopener" }
+    rel="noreferrer noopener"
```
kyle-bowden, line 25: The twirl template was not aligned with the hmrcfrontend nunjucks template. In the nunjucks template, the rel attribute is not optional.

oscarduignan, line 25: ah with this one, we want to do it the other way, and apply this change to the hmrc-frontend version- I think this was a quick fix done directly to play-frontend-hmrc because of a live issue that got missed from hmrc-frontend for expediency and I forgot to go back after and update hmrc-frontend 🤦

oscarduignan:
other than that it looks good to me 👍

src/main/twirl/uk/gov/hmrc/hmrcfrontend/views/components/HmrcReportTechnicalIssue.scala.html
kyle-bowden: That makes sense, I actually did the change in hmrc-frontend and then thought that was the wrong approach. I will resurface that pull request then.

oscarduignan: 🙌

PR #190, opened 2022-09-21
PLATUI-1907 spike possible API for enabling new Javascript back link helper in hmrc-frontend

ellamdav:


src/main/scala/uk/gov/hmrc/hmrcfrontend/config/BackLinkConfig.scala
```diff
@@ -0,0 +1,26 @@
...
+
+package uk.gov.hmrc.hmrcfrontend.config
+
+import play.api.Configuration
+
+import javax.inject.Inject
+
+class BackLinkConfig @Inject() (configuration: Configuration) {
+  def enableBrowserBack: Boolean =
+    configuration.getOptional[Boolean]("hmrc-back-link.enable-browser-back-if-no-url").getOrElse(true)
```
ellamdav, line 25: `getOrElse(true)` implies it's on by default, to ease adoption, but may break services already rolling their own solutions unless they add config to switch off (or migrate to our solution)

src/main/scala/uk/gov/hmrc/hmrcfrontend/views/Utils.scala
```diff
@@ -24,6 +25,12 @@ trait Utils extends UtilsSupport {
     path
       .map(p => s"$p/$file")
       .getOrElse(uk.gov.hmrc.hmrcfrontend.controllers.routes.Assets.at(file).url)
+
+  private[views] def maybeBackLinkDataModule(href: String, backLinkConfig: BackLinkConfig): Option[String] =
+    (href match {
+      case unset if unset == "" || unset == "#" => Some(""" data-module="hmrc-back-link"""")
+      case _                                    => None
+    }).filter(_ => backLinkConfig.enableBrowserBack)
```
ellamdav, line 33: basically, if the `href` is `#` or blank, and the feature is enabled, add the `data-module` attribute to trigger the JS in hmrc-frontend

src/main/scala/uk/gov/hmrc/govukfrontend/views/viewmodels/backlink/BackLink.scala
```diff
@@ -32,6 +32,8 @@ object BackLink {
 
   def defaultObject: BackLink = BackLink()
 
+  def USE_BROWSER_BACK = "#"
```
ellamdav, line 35: This is just a hack that would allow services using the `hmrcLayout` to pass a dummy URL that will trigger the JS, like
```
@hmrcLayout(backLinkUrl = Some(BackLink.USE_BROWSER_BACK))
```
because passing `None` means the back link isn't rendered at all.  Probably needs modelling properly, but in a way that won't break existing services.

oscarduignan: I reckon in practice we might want to add another attribute like `backLink=Some(BackLink)` then I could see rather than using the a special url value, have something like this:

```
object BackLink {
   ...   

   val mimickBrowserBackButtonViaJavaScript = BackLink(
     href = "#",
     attributes = "data-module" -> "hmrc-back-link"
   )

   ...
}
```

src/main/scala/uk/gov/hmrc/govukfrontend/views/viewmodels/backlink/BackLink.scala
```diff
@@ -32,6 +33,12 @@ object BackLink {
 
   def defaultObject: BackLink = BackLink()
 
+  def mimicsBrowserBackButtonViaJavaScript(implicit messages: Messages): BackLink = BackLink(
```
ellamdav, line 36: I'm assuming this would generally be used in a context where `Messages` are available, and we'd want to avoid having to explicitly pass in `content`?

src/test/scala/uk/gov/hmrc/hmrcfrontend/views/helpers/hmrcLayoutSpec.scala
```diff
@@ -438,6 +439,55 @@ class hmrcLayoutSpec extends AnyWordSpecLike with Matchers with JsoupHelpers wit
...
+
+      val actualBackLink = document.select(".govuk-back-link")
+      actualBackLink                            should have size 1
+      actualBackLink.attr("href")             shouldBe "/some/url"
+      actualBackLink.hasClass("custom-class") shouldBe true
+      actualBackLink.attr("some")             shouldBe "attribute"
+      actualBackLink.html()                   shouldBe "Custom Back"
+    }
+
+    "add attributes to activate the JavaScript browser back in hmrc-frontend" in {
```
ellamdav, line 459: this is really a unit test of `BackLink.mimicsBrowserBackButtonViaJavaScript` but feels more coherent/discoverable to have here?

src/test/scala/uk/gov/hmrc/hmrcfrontend/views/helpers/hmrcLayoutSpec.scala
```diff
@@ -438,6 +439,55 @@ class hmrcLayoutSpec extends AnyWordSpecLike with Matchers with JsoupHelpers wit
...
+      val document                    = Jsoup.parse(contentAsString(layout))
+
+      val actualBackLink = document.select(".govuk-back-link")
+      actualBackLink                       should have size 1
+      actualBackLink.attr("href")        shouldBe "#"
+      actualBackLink.attr("data-module") shouldBe "hmrc-back-link"
+      actualBackLink.html()              shouldBe "Back English"
+    }
+
+    "give backLink precedence over backLinkUrl" in {
```
ellamdav, line 474: I'm assuming this is what we want, rather than throwing an exception or showing both?

src/main/scala/uk/gov/hmrc/govukfrontend/views/viewmodels/backlink/BackLink.scala
oscarduignan: this sounds good to me, doing it this way matches the [current default](https://github.com/hmrc/play-frontend-hmrc/blob/main/src/main/twirl/uk/gov/hmrc/hmrcfrontend/views/helpers/HmrcLayout.scala.html#L87) we use, and if they want to override something they can `.copy()` it

oscarduignan: I'm somewhat assuming with this that people won't want to have a fallback url as well, because if they have a real url to go back to then it would be better to not use this js behaviour anyway

src/main/scala/uk/gov/hmrc/hmrcfrontend/config/BackLinkConfig.scala
ellamdav: removed this now

src/main/scala/uk/gov/hmrc/hmrcfrontend/views/Utils.scala
ellamdav: removed this now

src/test/scala/uk/gov/hmrc/hmrcfrontend/views/helpers/hmrcLayoutSpec.scala
oscarduignan: sorry for the delay, meetings interrupted me. Maybe we should @require that we only get one or the other, I could imagine that people might pass both by mistake and have a wrong assumption about how it will behave that they don't notice. Since backLink is new it doesn't feel dangerous to do, and we've used it in other components to guard against stuff https://github.com/hmrc/play-frontend-hmrc/blob/8969a6a0a56d1b185b9d67bfa603b07f5bd28170/src/main/twirl/uk/gov/hmrc/govukfrontend/views/components/GovukCharacterCount.scala.html#L23

oscarduignan (dismissed):

oscarduignan (approved):

PR #192, opened 2022-10-03
PLATUI-1830: Productionize the accessible autocomplete component in play-frontend-hmrc

kyle-bowden:
There will be another ticket to come out of this for tidying up the HmrcLayout parameter list, it has reached its max parameter count of 22. I was unable to add the `includeAccessibleAutocompleteAssets` property proposed in this proposal [document](https://docs.google.com/document/d/1MmrYmPnoJ9M8ZdyAe3LC6oxdpknZdJo6MjxnhoguRPg) because of this.


src/main/scala/uk/gov/hmrc/govukfrontend/views/implicits/RichSelectSupport.scala
```diff
@@ -61,6 +62,22 @@ trait RichSelectSupport {
...
+          "data-show-all-values" -> accessibleAutocomplete.showAllValues.toString,
+          "data-default-value"   -> accessibleAutocomplete.defaultValue.getOrElse(""),
+          "data-module"          -> accessibleAutocomplete.dataModule
+        )
+
+      val dataAttributes =
+        toMapOfDataAttributes(accessibleAutocomplete.getOrElse(AccessibleAutocomplete(None))) ++
+          select.attributes
+
+      select.copy(attributes = dataAttributes)
```
ellamdav, line 78: think this would be slightly clearer as
```suggestion
      val dataAttributes =
        toMapOfDataAttributes(accessibleAutocomplete.getOrElse(AccessibleAutocomplete(None)))

      select.copy(attributes = select.attributes ++ dataAttributes)
```

src/main/scala/uk/gov/hmrc/hmrcfrontend/views/viewmodels/accessibleautocomplete/AccessibleAutocomplete.scala
```diff
@@ -0,0 +1,33 @@
...
+  defaultValue: Option[String] =
+    None, // Please read on the usage of the defaultValue property at https://www.npmjs.com/package/accessible-autocomplete under the `Null options` heading
+  showAllValues: Boolean = false,
+  autoSelect: Boolean = false
+) {
+  val dataModule: String = "hmrc-accessible-autocomplete"
+}
+
+object AccessibleAutocomplete {
+  implicit def jsonFormats: OFormat[AccessibleAutocomplete] =
```
ellamdav, line 31: is this needed?  just wondering where this would be (de)serialised.

src/test/scala/uk/gov/hmrc/govukfrontend/views/implicits/RichSelectSpec.scala
```diff
@@ -165,4 +166,50 @@ class RichSelectSpec extends AnyWordSpec with Matchers with MessagesHelpers with
...
+  "Given a Select object, calling asAccessibleAutocomplete" should {
+    "set the accessible autocomplete data attributes" in {
+      val select        = Select()
+      val updatedSelect = select.asAccessibleAutocomplete(Some(AccessibleAutocomplete(Some("Test"))))
+
+      updatedSelect shouldBe Select(attributes =
+        Map(
+          "data-default-value"   -> "Test",
+          "data-show-all-values" -> "false",
+          "data-auto-select"     -> "false",
```
ellamdav, line 179: worth testing non-default values for the other two attributes here?
```suggestion
      val updatedSelect = select.asAccessibleAutocomplete(Some(AccessibleAutocomplete(Some("Test"), true, true)))

      updatedSelect shouldBe Select(attributes =
        Map(
          "data-default-value"   -> "Test",
          "data-show-all-values" -> "true",
          "data-auto-select"     -> "true",
```

src/test/scala/uk/gov/hmrc/hmrcfrontend/views/TemplateUnitBaseSpec.scala
```diff
@@ -33,7 +33,18 @@ abstract class TemplateUnitBaseSpec[T: Reads](hmrcComponentName: String)
     "character-count-spellcheck-enabled"
   )
 
-  override protected val skippedExamples: Seq[String] = skipBecauseOfSpellcheckOrdering
+  val skipBecauseNoActualComponents = Seq(
```
ellamdav, line 36: 👍

ellamdav:
> There will be another ticket to come out of this for tidying up the HmrcLayout parameter list, it has reached its max parameter count of 22. I was unable to add the `includeAccessibleAutocompleteAssets` property proposed in this proposal [document](https://docs.google.com/document/d/1MmrYmPnoJ9M8ZdyAe3LC6oxdpknZdJo6MjxnhoguRPg) because of this.

😬

README.md
```diff
@@ -847,6 +847,95 @@ You will need to inject the CSS and Javascript into your views as follows:
...
+showAllValues = false
+autoselect = false
+```
+More information on these values can be found [here](https://www.npmjs.com/package/accessible-autocomplete) under the `API documentation` heading
+
+If you wish to change these values you can do so by providing an `AccessibleAutocomplete` model into the `asAccessibleAutocomplete` method as seen below
+```scala
+@govukSelect(Select(...).asAccessibleAutocomplete(Some(
+  AccessibleAutocomplete(
+    defaultValue = Some("United Kingdom"),
```
oscarduignan, line 872: one thing I thought, is I bet this is so that you can have an initial "placeholder text" that makes sense when it's a normal select "select a country", then when it's enhanced you might want to adjust it to "start typing to select a country" 🤷‍♂️ just `defaultValue=` attribute name for me doesn't convey that kind of thing at all

ellamdav: yeah, it seems this is really `defaultText`, but i guess we're just mirroring the alphagov API

kyle-bowden: It is misleading, should we change the AccessibleAutocomplete model to have defaultText instead or promptText?

oscarduignan: @kyle-bowden maybe rather than "United Kingdom" in the example we could put "Start typing to select a country" and in the bit below have "Placeholder text" be "Select an option" - copy probably not quite inline with GDS guidance but makes it clear that this is all around placeholder type text

oscarduignan: or actually, I'm dithering, looking at the guidance that's not how they use it... lets be consistent with GDS

oscarduignan (approved):

PR #198, opened 2022-11-30
PLATUI-2089: Uplift to use govuk-frontend v4.4.0

kyle-bowden:


src/it/scala/uk/gov/hmrc/govukfrontend/views/components/GovukCharacterCountIntegrationSpec.scala
```diff
@@ -22,7 +22,8 @@ import uk.gov.hmrc.govukfrontend.views.viewmodels.charactercount.CharacterCount
 import uk.gov.hmrc.govukfrontend.views.viewmodels.charactercount.Generators._
 
 object GovukCharacterCountIntegrationSpec
-    extends TemplateIntegrationSpec[CharacterCount, GovukCharacterCount](
-      govukComponentName = "govukCharacterCount",
-      seed = None
-    )
+// TODO: awaiting decision from ticket https://jira.tools.tax.service.gov.uk/browse/PLATUI-2117
```
kyle-bowden, line 25: I have had to disable integration tests for the character count component as the difference between the nunjucks and twirl versions is too different to try to fix without making significant changes to the twirl component.

src/main/twirl/uk/gov/hmrc/govukfrontend/views/components/GovukHeader.scala.html
```diff
@@ -79,8 +79,8 @@
           }
 
           @navigation.map { navigation =>
-            <nav aria-label="@params.navigationLabel.getOrElse("Menu")" class="govuk-header__navigation @navigationClasses">
-              <button type="button" class="govuk-header__menu-button govuk-js-header-toggle" aria-controls="navigation" aria-label="@params.menuButtonLabel.getOrElse("Show or hide menu")" hidden>Menu</button>
+            <nav aria-label="@params.navigationLabel.getOrElse(menuButtonText.getOrElse("Menu"))" class="govuk-header__navigation @navigationClasses">
```
kyle-bowden, line 82: I done the work here already but I can always strip it out as we have a ticket crated for this work
this was in the ticket https://jira.tools.tax.service.gov.uk/browse/PLATUI-2116
It just felt like a trivial change.

project/GenerateFixtures.scala
```diff
@@ -43,13 +43,16 @@ case class GenerateFixtures(fixturesDir: File, frontend: String, version: String
     def patchedVersionDoesNotExist(example: JsObject): Boolean =
       !(patchedFixturesDir / (example \ "exampleId").as[String]).isDirectory
 
+    def replaceUnwantedCharacters(text: String) = text.replaceAll("/", "-")
```
kyle-bowden, line 46: We have a problem when generating example folders from the examples provided by the govuk library
some example names have special characters in them. In this case the one example contained a / in the name

`when neither maxlength/maxwords nor textarea description are set`

But I have seen a few example folders with other special characters like below

`radios-with-fieldset,-error-message-and-describedBy
notification-banner-role=alert-overridden-to-role=region,-with-type-as-success
checkboxes-with-single-option-(and-hint)-set-'aria-describedby'-on-input,-and-describedBy`

I will need another ticket to sort this issue out.

src/it/scala/uk/gov/hmrc/govukfrontend/views/components/GovukCharacterCountIntegrationSpec.scala
oscarduignan: I think in this release we _**should**_ update the govuk components, but skip porting them changes to the forked hmrc versions to separate tickets

src/main/scala/uk/gov/hmrc/govukfrontend/views/Utils.scala
```diff
@@ -30,6 +34,10 @@ trait Utils extends UtilsSupport {
       .map(p => s"$p/$file")
       .getOrElse(uk.gov.hmrc.hmrcfrontend.controllers.routes.Assets.at(s"govuk/$file").url)
 
+  private[views] def generateI18TranslationAttributes(strings: Option[TranslatableStrings], attributeName: String) =
+    strings.getOrElse(Map.empty).map { case (k, v) =>
+      Html(s"""$i18nDataAttribute.$attributeName.${HtmlFormat.escape(k)}="${HtmlFormat.escape(v)}" """)
+    }
```
oscarduignan, line 40: @kyle-bowden could we copy the name / params from https://github.com/alphagov/govuk-frontend/blob/d92a05aeea22236c00760feb5289f450ab9035c6/src/govuk/macros/i18n.njk#L13-L15 and I reckon rather than having a private i18nDataAttribute just have the data-i18n inline below because it's used once

src/main/scala/uk/gov/hmrc/govukfrontend/views/viewmodels/charactercount/CharacterCount.scala
```diff
@@ -38,11 +39,20 @@ case class CharacterCount(
...
+  charactersOverLimitText: Option[TranslatableStrings] = None,
+  wordsUnderLimitText: Option[TranslatableStrings] = None,
+  wordsAtLimitText: Option[String] = None,
+  wordsOverLimitText: Option[TranslatableStrings] = None,
+  textareaDescriptionText: Option[String] = None
 )
 
 object CharacterCount {
 
+  type TranslatableStrings = Map[String, String]
```
oscarduignan, line 54: Could we move and rename this to `uk.gov.hmrc.govukfrontend.views.viewmodels.i18n.PluralForms` to copy the name from https://github.com/alphagov/govuk-frontend/blob/d92a05aeea22236c00760feb5289f450ab9035c6/src/govuk/macros/i18n.njk#L13-L15 and because I think this will be reused

oscarduignan: guess we'll also need to add a type alias for it https://github.com/hmrc/play-frontend-hmrc/blob/273ef63bedf6cfc91efb82cee8a9b83d9e18f8a5/src/main/scala/uk/gov/hmrc/govukfrontend/views/Aliases.scala can you have a type alias pointing to another type alias?

src/main/twirl/uk/gov/hmrc/govukfrontend/views/components/GovukButton.scala.html
```diff
@@ -41,8 +41,10 @@
 
 @commonAttributes(classNames: String) = {class="@classNames" data-module="govuk-button"@toAttributes(attributes)}
 
+@hasPreventDoubleClick = @{preventDoubleClick.isDefined}
+
 @buttonAttributes = {
-  @name.mapNonEmpty {name => name="@name"}@if(disabled) { disabled="disabled" aria-disabled="true"}@if(preventDoubleClick) { data-prevent-double-click="true"}
+  @name.mapNonEmpty {name => name="@name"}@if(disabled) { disabled="disabled" aria-disabled="true"}@if(hasPreventDoubleClick) { data-prevent-double-click="@{preventDoubleClick.get}"}
```
oscarduignan, line 47: this might work, though have to check that it adds spaces before attribute name correctly. As an aside it's interesting to see that previously if it was `false` then we wouldn't output any attribute and I guess it would use the default, where as now if it's set to false we'll set an attribute. Seems like the old behaviour could have meant (if default ever changed to true) that there was no way to turn it off 😅 
```suggestion
  @name.mapNonEmpty {name => name="@name"}@if(disabled) { disabled="disabled" aria-disabled="true"}@preventDoubleClick.map {value => data-prevent-double-click="@value"}
```

src/main/twirl/uk/gov/hmrc/govukfrontend/views/components/GovukNotificationBanner.scala.html
```diff
@@ -39,10 +39,12 @@
     titleId.getOrElse("govuk-notification-banner-title")
 }
 
+@hasDisableAutoFocus = @{disableAutoFocus.isDefined}
+
 <div class="@toClasses("govuk-notification-banner", classes, if(bannerType.contains("success"))" govuk-notification-banner--success" else "")"
     role="@{bannerRole}"
     aria-labelledby="@{headingAndTitleId}"
-    data-module="govuk-notification-banner"@if(disableAutoFocus.getOrElse(false)){ data-disable-auto-focus="true"}@toAttributes(attributes)>
+    data-module="govuk-notification-banner"@if(hasDisableAutoFocus){ data-disable-auto-focus="@{disableAutoFocus.get}"}@toAttributes(attributes)>
```
oscarduignan, line 47: same as before, could it be a map instead, because twirl will automatically unwrap options https://www.playframework.com/documentation/2.8.x/ScalaTemplates#Displaying-Scala-types

src/main/scala/uk/gov/hmrc/govukfrontend/views/viewmodels/charactercount/CharacterCount.scala
kyle-bowden: This is looking quite odd 😆 I have an object 

`package uk.gov.hmrc.govukfrontend.views.viewmodels.i18n

object PluralForms {
  type PluralForms = Map[String, String]
}`

Then I have an Alias in Aliases pointing to the type in the object

`import uk.gov.hmrc.govukfrontend.views.viewmodels.i18n.PluralForms

trait Aliases {
  type PluralForms = PluralForms.PluralForms`

It is happy with this

src/main/scala/uk/gov/hmrc/govukfrontend/views/viewmodels/errorsummary/ErrorSummary.scala
```diff
@@ -26,7 +26,7 @@ case class ErrorSummary(
   attributes: Map[String, String] = Map.empty,
   title: Content = Empty,
   description: Content = Empty,
-  disableAutoFocus: Boolean = false
+  disableAutoFocus: Option[Boolean] = Some(false)
```
oscarduignan, line 29: @kyle-bowden just realising with my Friday brain that where existing params have changed from Boolean to Option[Boolean] is a breaking change, wonder if we should go back to what we had before and just have a slight discrepancy with govuk API for backwards compatibility since I checked and [default is still false in the javascript](https://github.com/alphagov/govuk-frontend/blob/main/src/govuk/components/error-summary/error-summary.mjs#L33) - or if we should do a major version bump and describe the changes needed changelog

src/main/scala/uk/gov/hmrc/govukfrontend/views/viewmodels/i18n/PluralForms.scala
```diff
@@ -0,0 +1,21 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package uk.gov.hmrc.govukfrontend.views.viewmodels.i18n
+
+object PluralForms {
+  type PluralForms = Map[String, String]
```
oscarduignan, line 20: @kyle-bowden I know I was the one who mentioned this, but actually seeing it I'm realising it's providing no value - think we can just remove this and use directly `Map[String, String]` just my brain trying to overcomplicate stuff

kyle-bowden: I do like the idea of giving types like this, as it makes it easier to understand what's happening but I think in this instance its probably just easier to keep it as a Map[String, String]

oscarduignan (approved):

PR #206, opened 2022-12-21
PLATUI-1987: Add nvmrc file

kyle-bowden:


oscarduignan (approved):
still really unsure if this will trigger a new version 😅 wish there was an indication on PR and a test for PRs where if it will trigger a new version and there isn't a new version it tells us

PR #210, opened 2023-01-05
PLATUI-2149 add value parameter to Radios, Checkboxes and Select components

ellamdav:
* add `values` param to `Checkboxes` - this takes precedence over any `checked` `items`
* add `value` param to `Radios` - this takes precedence over any `checked` `item`
* add `value` param to `Select` - this takes precedence over any `selected` `items`
* add anti-corruption checks in these templates to protect against illogical state
* introduce excluded-fixtures directory to replace `skippedExamples`, and documented existing excluded fixtures
* mark excluded fixtures as pending in test output, instead of silently ignoring
* use tolerant JSON reader to fix existing skipped fixture where `value` failed JSON validation


src/test/scala/uk/gov/hmrc/govukfrontend/views/components/GovukRadiosSpec.scala
```diff
@@ -19,4 +19,34 @@ package components
 
 import uk.gov.hmrc.govukfrontend.views.html.components._
 
-class GovukRadiosSpec extends TemplateUnitSpec[Radios, GovukRadios]("govukRadios")
+class GovukRadiosSpec extends TemplateUnitSpec[Radios, GovukRadios]("govukRadios") {
+
+  "anti-corruption layer" should {
+    "throw if more than one radio item is checked" in {
```
oscarduignan, line 25: I did a quick browser test and it seems like the last one with checked attribute ends up selected when there are multiple. I think generally it's better to blow up to avoid wrong behaviour, but the way it could go wrong without the require() asking for only one checked item is that the wrong default value could be pre-checked but the user could correct and continue from that, where with the require() we're introducing an unrecoverable error that might prevent people doing stuff. I'm still tempted to say to do it, because otherwise bug might never be found in services if it actually exists anywhere, but I think this would need to be done as a breaking change with a note about it in the changelog

CleanShot 2023-01-06 at 09 09 10@2x

oscarduignan:
makes it a lot clearer having the README with the skipped fixtures 🙌 already highlights some other things to look at as part of this epic

src/test/scala/uk/gov/hmrc/govukfrontend/views/components/GovukCheckboxesSpec.scala
```diff
@@ -17,6 +17,48 @@
...
+        ),
+        values = Set("ketchup", "mayo")
+      )
+
+      the[IllegalArgumentException] thrownBy {
+        component.render(invalidCheckboxes)
+      } should have message "requirement failed: checked item(s) conflict with passed value(s)"
+    }
+
+    "not throw if passed values are ordered differently to the checked items" in {
```
ellamdav, line 49: we wrote this test and then decided that `values` should probably be a `Set` - not sure if this test really makes sense any more

oscarduignan (approved):

PR #212, opened 2023-01-06
PLATUI-2152: Update the documentation to make it harder to miss new components when updating the gov.uk frontend version

kyle-bowden:


docs/maintainers/upgrading.md
```diff
@@ -1,5 +1,55 @@
 # Upgrading play-frontend-hmrc
 
+## Warning
+
+When you bump the govuk-frontend version, you need to manually check release changes for new components and new parameters that need to be added to our twirl implementations. They won't currently be automatically detected as missing by our automated tests. Read [comparing differences](#comparing-differences) below for ways you can do this.
```
oscarduignan, line 5: If we format it like this:
```suggestion
> **Warning**
> When you bump the govuk-frontend version, you need to manually check release changes for new components and new parameters that need to be added to our twirl implementations. They won't currently be automatically detected as missing by our automated tests. Read [comparing differences](#comparing-differences) below for ways you can do this.
```
then it will get some fancy styling like this https://github.com/community/community/discussions/16925
CleanShot 2023-01-06 at 13 08 03@2x

oscarduignan: just to differentiate it a bit from the other writing on the page

kyle-bowden: Great, i have learned something new here today

oscarduignan (dismissed):

oscarduignan (approved):

PR #216, opened 2023-01-19
PLATUI-2010 create new HmrcStandardLayout to supersede HmrcLayout

ellamdav:
* create new `HmrcStandardPage` template that takes a single parameter `HmrcStandardPageParams`
* `HmrcStandardPageParams` encapsulates parameters from `HmrcLayout` template, grouped into logical entities:
  * `ServiceURLs` - service-specific URLs that can probably be set once
  * `Banners` - parameters relating to what banners to show on a page - probably set once but might be contextual
  * `TemplateOverrides` - parameters that override the default layout or inject additional HTML
  * `serviceName`, `pageTitle`, `isWelshTranslationAvailable`, `backLink` and `cspNonce` left at the top level - will be wired up in the service team's own `Layout` class
* Existing `HmrcLayout` now just wraps `HmrcStandardPage` - also provides a pattern for teams to easily migrate to the new template, if and when they need to
* new page params require typed `BackLink` and `UserResearchBanner` rather than URL strings
* TODO - how to mark old layout class as deprecated?

### DIFF
--- a/src/main/twirl/uk/gov/hmrc/hmrcfrontend/views/helpers/HmrcLayout.scala.html
+++ b/src/main/twirl/uk/gov/hmrc/hmrcfrontend/views/helpers/HmrcLayout.scala.html
@@ -14,14 +14,8 @@
  * limitations under the License.
  *@
 
-@import uk.gov.hmrc.govukfrontend.views.html.components.GovukBackLink
 @import uk.gov.hmrc.govukfrontend.views.html.components.TwoThirdsMainContent
-@import uk.gov.hmrc.hmrcfrontend.views.html.helpers.HmrcStandardHeader
-@import uk.gov.hmrc.hmrcfrontend.views.html.helpers.HmrcStandardFooter
 @import uk.gov.hmrc.hmrcfrontend.views.html.helpers.HmrcStandardPage
-@import uk.gov.hmrc.hmrcfrontend.views.html.helpers.HmrcHead
-@import uk.gov.hmrc.hmrcfrontend.views.html.helpers.HmrcLanguageSelectHelper
-@import uk.gov.hmrc.hmrcfrontend.views.html.helpers.HmrcScripts
 @import uk.gov.hmrc.hmrcfrontend.views.viewmodels.userresearchbanner.UserResearchBanner
 @import uk.gov.hmrc.govukfrontend.views.viewmodels.backlink.BackLink
 @import uk.gov.hmrc.govukfrontend.views.viewmodels.phasebanner.PhaseBanner
@@ -32,12 +26,6 @@
 
 @this(
         hmrcStandardPage: HmrcStandardPage,
-        hmrcStandardHeader: HmrcStandardHeader,
-        hmrcStandardFooter: HmrcStandardFooter,
-        hmrcHead: HmrcHead,
-        hmrcLanguageSelectHelper: HmrcLanguageSelectHelper,
-        hmrcScripts: HmrcScripts,
-        govukBackLink: GovukBackLink,
         defaultMainContent: TwoThirdsMainContent,
         fixedWidthPageLayout: FixedWidthPageLayout
 )
--- a/src/main/twirl/uk/gov/hmrc/hmrcfrontend/views/helpers/HmrcLayout.scala.html
+++ b/src/main/twirl/uk/gov/hmrc/hmrcfrontend/views/helpers/HmrcLayout.scala.html
@@ -14,24 +14,24 @@
  * limitations under the License.
  *@
 
-@import uk.gov.hmrc.govukfrontend.views.html.components.GovukLayout
 @import uk.gov.hmrc.govukfrontend.views.html.components.GovukBackLink
 @import uk.gov.hmrc.govukfrontend.views.html.components.TwoThirdsMainContent
 @import uk.gov.hmrc.hmrcfrontend.views.html.helpers.HmrcStandardHeader
 @import uk.gov.hmrc.hmrcfrontend.views.html.helpers.HmrcStandardFooter
+@import uk.gov.hmrc.hmrcfrontend.views.html.helpers.HmrcStandardPage
 @import uk.gov.hmrc.hmrcfrontend.views.html.helpers.HmrcHead
 @import uk.gov.hmrc.hmrcfrontend.views.html.helpers.HmrcLanguageSelectHelper
 @import uk.gov.hmrc.hmrcfrontend.views.html.helpers.HmrcScripts
 @import uk.gov.hmrc.hmrcfrontend.views.viewmodels.userresearchbanner.UserResearchBanner
 @import uk.gov.hmrc.govukfrontend.views.viewmodels.backlink.BackLink
 @import uk.gov.hmrc.govukfrontend.views.viewmodels.phasebanner.PhaseBanner
-@import uk.gov.hmrc.govukfrontend.views.viewmodels.content.Text
 @import uk.gov.hmrc.govukfrontend.views.viewmodels.pagelayout.PageLayout
+@import uk.gov.hmrc.hmrcfrontend.views.viewmodels.hmrcstandardpage._
 @import uk.gov.hmrc.govukfrontend.views.html.components.FixedWidthPageLayout
 @import uk.gov.hmrc.hmrcfrontend.views.Aliases.Header
 
 @this(
-        govukLayout: GovukLayout,
+        hmrcStandardPage: HmrcStandardPage,
         hmrcStandardHeader: HmrcStandardHeader,
         hmrcStandardFooter: HmrcStandardFooter,
         hmrcHead: HmrcHead,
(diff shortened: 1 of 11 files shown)


src/main/scala/uk/gov/hmrc/govukfrontend/views/viewmodels/layout/HmrcLayoutParams.scala
```diff
@@ -0,0 +1,29 @@
...
+ * limitations under the License.
+ */
+
+package uk.gov.hmrc.govukfrontend.views.viewmodels.layout
+
+import uk.gov.hmrc.govukfrontend.views.Aliases.BackLink
+
+// TODO name?
+case class HmrcLayoutParams(
+  pageTitle: Option[String] = None,
```
kyle-bowden, line 23: I'm just thinking maybe we should have value classes for **pageTitle** and **nonce** to keep all parameters typed while we have the opportunity during this refactor

src/main/scala/uk/gov/hmrc/govukfrontend/views/viewmodels/layout/Service.scala
```diff
@@ -0,0 +1,25 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package uk.gov.hmrc.govukfrontend.views.viewmodels.layout
+
+case class Service(
+  serviceName: Option[String] = None,
```
kyle-bowden, line 20: Same as my comment above

kyle-bowden:
**TODO - agree structure / naming of new layout class**
I personally believe the structure you have now is well-defined as for the name of the layout I just cant think of anything good.

**TODO - how to mark old layout class as deprecated?**
I believe we can use scala standard deprecated annotation in a template like

`@deprecated("This template is deprecated and will be removed in a future version", "X.X.X")`

Once the template is marked as deprecated, the compiler will generate a warning when the template is used and the message passed on the annotation will be shown

ellamdav:
> **TODO - how to mark old layout class as deprecated?** I believe we can use scala standard deprecated annotation in a template like
> 
> `@deprecated("This template is deprecated and will be removed in a future version", "X.X.X")`
> 
> Once the template is marked as deprecated, the compiler will generate a warning when the template is used and the message passed on the annotation will be shown
I've tried this, but annotating the view itself doesn't compile.  Also tried annotating the `contentBlock` parameter but, although this compiles, I didn't get any deprecation warning when compiling a consuming service.

src/main/scala/uk/gov/hmrc/govukfrontend/views/viewmodels/layout/HmrcLayoutParams.scala
ellamdav: > **TODO - agree structure / naming of new layout class** I personally believe the structure you have now is well-defined as for the name of the layout I just cant think of anything good.

maybe it should be `uk.gov.hmrc.govukfrontend.views.viewmodels.hmrcstandardlayout.Params` to make it clear it's not a hmrc-frontend component but models for the helper?

ellamdav:
> **TODO - agree structure / naming of new layout class** I personally believe the structure you have now is well-defined as for the name of the layout I just cant think of anything good.

maybe it should be `uk.gov.hmrc.govukfrontend.views.viewmodels.hmrcstandardlayout.Params` to make it clear it's not a hmrc-frontend component but models for the helper?

kyle-bowden:
> > **TODO - agree structure / naming of new layout class** I personally believe the structure you have now is well-defined as for the name of the layout I just cant think of anything good.
> 
> maybe it should be `uk.gov.hmrc.govukfrontend.views.viewmodels.hmrcstandardlayout.Params` to make it clear it's not a hmrc-frontend component but models for the helper?

That sounds sensible to me.

src/main/scala/uk/gov/hmrc/hmrcfrontend/views/viewmodels/layout/HmrcStandardLayoutParams.scala
```diff
@@ -0,0 +1,28 @@
...
+
+import uk.gov.hmrc.govukfrontend.views.Aliases.BackLink
+
+case class HmrcStandardLayoutParams(
+  service: Service = Service(),
+  banners: Banners = Banners(),
+  customLayout: CustomLayout = CustomLayout(),
+  pageTitle: Option[String] = None,
+  backLink: Option[BackLink] = None,
+  cspNonce: Option[String] = None
```
oscarduignan, line 27: the names for the bits of this view model are quite generic, because we have a convention of [hoisting stuff into the Aliases package](https://github.com/hmrc/play-frontend-hmrc/blob/main/src/main/scala/uk/gov/hmrc/hmrcfrontend/views/Aliases.scala#LL19) so that people can wildcard import all the viewmodels - I wonder if having Banners / Service / CustomLayout will cause an issue with that convention or having them generic makes collisions more likely. Though equally I don't like the idea of having long shared prefix "HmrcStandardLayoutParamsService" - the way that aliases stuff works feels like another ticket that might be good to play to see if we can simplify it 🤷‍♂️

src/main/scala/uk/gov/hmrc/hmrcfrontend/views/viewmodels/layout/CustomLayout.scala
```diff
@@ -0,0 +1,34 @@
...
+import uk.gov.hmrc.hmrcfrontend.views.Aliases.Header
+
+case class CustomLayout(
+  additionalHeadBlock: Option[Html] = None,
+  additionalScriptsBlock: Option[Html] = None,
+  beforeContentBlock: Option[Html] = None,
+  mainContentLayout: Option[Html => Html] = None,
+  pageLayout: Option[PageLayout => Html] = None,
+  headerContainerClasses: String = Header.defaultObject.containerClasses
+)
```
oscarduignan, line 30: @ellamdav rather than CustomLayout, what do you think about something like LayoutOverrides or TemplateOverrides? Initially I read it as "no main content layout" and "no page layout" by default, rather than you're not overriding the default (where the parameters before had the default values in the case class) so the name of the params are the same but the semantics of it's value are now different

src/main/twirl/uk/gov/hmrc/hmrcfrontend/views/helpers/HmrcLayout.scala.html
```diff
@@ -62,44 +65,35 @@
...
-        @if(isWelshTranslationAvailable) {
-            @hmrcLanguageSelectHelper()
-        }
-        @{(backLink orElse backLinkUrl.map(url => BackLink(href = url, content = Text(messages("back.text")))))
-            .map(link => govukBackLink(link))}
-    }
-}
-
-@govukLayout(
+@hmrcStandardLayout(
```
oscarduignan, line 72: @ellamdav what do you think about HmrcStandardPage rather than HmrcStandardLayout - looks really good to me otherwise, feels like service / banners / layout / and page content at the top level is a good way to shuffle things

src/main/scala/uk/gov/hmrc/hmrcfrontend/views/viewmodels/layout/HmrcStandardLayoutParams.scala
ellamdav: I'm now wondering if I should ditch this `HmrcStandardLayoutParams` and just pass its contents as parameters to the new template.  Not sure it's adding much, other than more boilerplate in the code & tests.

ellamdav: Agree that `Service` and `Banners` are really generic...  and actually, on closer inspection `serviceName` is intended for use when you have a dynamic service name depending on context - by default the service name just comes from messages `service.name`.  So not sure it belongs here, should perhaps be `dynamicServiceName` alongside `pageTitle` etc.?

ellamdav: Maybe should have a container for `ServiceURLs` but leave the other stuff out?

src/main/scala/uk/gov/hmrc/hmrcfrontend/views/viewmodels/layout/CustomLayout.scala
ellamdav: Yeah, I'm not really happy with how this bit now works.  The defaults were previously set using injected templates, but we can't do that in a simple case class constructor, so I've ended up adding that awkward `CustomLayout.noMainContentLayout`.

src/main/scala/uk/gov/hmrc/hmrcfrontend/views/viewmodels/layout/HmrcStandardLayoutParams.scala
ellamdav: have replaced `Service` with `ServiceURLs`, with the non-URL fields just pulled up to the top level `...Params`

src/main/scala/uk/gov/hmrc/hmrcfrontend/views/viewmodels/layout/CustomLayout.scala
ellamdav: renamed `CustomLayout` to `TemplateOverrides`

src/main/twirl/uk/gov/hmrc/hmrcfrontend/views/helpers/HmrcLayout.scala.html
ellamdav: renamed to `HmrcStandardPage`

src/main/twirl/uk/gov/hmrc/hmrcfrontend/views/helpers/HmrcLayout.scala.html
```diff
@@ -14,24 +14,24 @@
...
 @import uk.gov.hmrc.govukfrontend.views.viewmodels.phasebanner.PhaseBanner
-@import uk.gov.hmrc.govukfrontend.views.viewmodels.content.Text
 @import uk.gov.hmrc.govukfrontend.views.viewmodels.pagelayout.PageLayout
+@import uk.gov.hmrc.hmrcfrontend.views.viewmodels.hmrcstandardpage._
 @import uk.gov.hmrc.govukfrontend.views.html.components.FixedWidthPageLayout
 @import uk.gov.hmrc.hmrcfrontend.views.Aliases.Header
 
 @this(
-        govukLayout: GovukLayout,
+        hmrcStandardPage: HmrcStandardPage,
```
oscarduignan, line 34: @ellamdav this PR looks good to go to me, the only bit I just spotted was that are some of these injected components no longer needed now? Other than that 🙌

ellamdav: thanks - another place where syntax highlighting in twirl templates could be improved 😅 .  Removed now.

oscarduignan (approved):

PR #217, opened 2023-01-25
NOJIRA try removing explicit CSP nonce parameters from templates

ellamdav:
and just pull it from the request


ellamdav:
Replaced by https://github.com/hmrc/play-frontend-hmrc/pull/226

PR #220, opened 2023-02-15
PLATUI-2116: Added translations for visible and screen reader menu te…

JoPintoPaul:
…xt in header


src/main/twirl/uk/gov/hmrc/hmrcfrontend/views/components/HmrcHeader.scala.html
```diff
@@ -77,10 +77,9 @@
         @if(serviceName.exists(_.nonEmpty) || signOutHref.exists(_.nonEmpty) || (navigation.getOrElse(Seq.empty).nonEmpty) || languageToggle.exists(_.linkMap.nonEmpty)) {
         <div class="govuk-header__content">
           @serviceNameLinkOrSpan(serviceName, serviceUrl)
-
           @if(navigation.getOrElse(Seq.empty).nonEmpty) {
-          <nav class="govuk-header__navigation" aria-label="Top Level Navigation">
-            <button type="button" class="govuk-header__menu-button govuk-js-header-toggle" aria-controls="navigation" aria-label="Show or hide Top Level Navigation" hidden>Menu</button>
+          <nav class="govuk-header__navigation" aria-label="@{if(language == Cy) "Gwelywio Lefel Uchaf" else "Top Level Navigation"}">
+            <button type="button" class="govuk-header__menu-button govuk-js-header-toggle" aria-controls="navigation" aria-label="@{if(language == Cy) {HtmlFormat.raw("Dangos neu guddio\'r Gwelywio Lefel Uchaf")} else {HtmlFormat.raw("Show or hide Top Level Navigation")}}" hidden>@{if(language == Cy) {HtmlFormat.escape("Dewislen")} else {HtmlFormat.escape("Menu")}}</button>
```
oscarduignan, line 82: @JoPintoPaul I can't see anything in here that would need escaping, is it doing a different kind of thing maybe because of the `@{if() "` rather than `@if() {}`
```suggestion
            <button type="button" class="govuk-header__menu-button govuk-js-header-toggle" aria-controls="navigation" aria-label="@if(language == Cy) {Dangos neu guddio'r Gwelywio Lefel Uchaf} else {Show or hide Top Level Navigation}" hidden>@if(language == Cy) {Dewislen} else {Menu}</button>
```

oscarduignan: I think @{if(something) "a" else "b"} would turn into `HtmlFormat.escape("a")` because of auto escaping of @ when outputting variables (and it's maybe doing scala inline to get a value and outputting that), whereas @if(something) { this will be be html } because it's not outputting a variable it's using a twirl control structure doesn't escape - might be totally off, just haven't seen the HtmlFormat used inline like that I think

JoPintoPaul: @oscarduignan The `HtmlFormat.raw` is needed for that apostrophe in the Welsh `Dangos neu guddio'r Gwelywio Lefel Uchaf`. Otherwise it's getting escaped when rendered.

The use of `HtmlFormat.escape`  was just based on the PlatOps blogpost advice to use `HtmFormat.escape` as a preference - but maybe it's overcomplicating it. 

Let me know thoughts.

oscarduignan: I think if we were using variables it could make sense to use `HtmlFormat.raw()` but I don't think we want to explicitly wrap all the blocks of inline html in our templates in `HtmlFormat.raw()`

oscarduignan (approved):

PR #221, opened 2023-02-17
PLATUI-2125: Refactored HmrcCharacterCount to use updated character-c…

JoPintoPaul:
…ount from govuk-frontend


src/main/twirl/uk/gov/hmrc/hmrcfrontend/views/components/HmrcCharacterCount.scala.html
```diff
@@ -14,55 +14,79 @@
...
  @if(messagesLanguage == Cy) {
-  @govukHint(Hint(
-   content = Text("Gallwch nodi hyd at " + params.maxLength.getOrElse(params.maxWords.getOrElse(0)) + {if(params.maxWords.isDefined) " o eiriau" else " o gymeriadau"}),
-   id = Some(params.id + "-info"),
-   classes = "hmrc-character-count__message" + {if (params.countMessageClasses == "") "" else " " + params.countMessageClasses}
-  ))
+  @defining(
+   messages match {
+    case messagesImpl: MessagesImpl => messagesImpl.messagesApi.preferred(Seq(Lang("cy")))
+    case _ => messages
```
oscarduignan, line 32: @JoPintoPaul what do you think about having the welsh translations inline rather than in the messages file and only using the messages implicit to check what the current language and maybe override the passed language param if it seems like they have sent En by mistake? We're adding the english translations to the messages file in this change but not using them

JoPintoPaul: @oscarduignan I did dither on this one! In other places we use translated strings inline, I'm not wild about that pattern but maybe it does make sense to be consistent. I'm hoping to go with group consensus on this one.

JoPintoPaul: @oscarduignan Yes the English translations are really there as a fallback only but I would struggle to think of a way they can be reached, not sure if the `Messages` would ever fall back to them as default.

oscarduignan: wish we could ignore params.language and just use the messages implicit with no branching - seems very unlikely that people would want a welsh character count but not have rest of the site using any welsh translations from messages

PR #222, opened 2023-02-20
PLATUI-2125: Refactored HmrcCharacterCount to use updated character-c…

JoPintoPaul:
…ount from govuk-frontend


CHANGELOG.md
```diff
@@ -8,7 +8,19 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 For compatibility information see `govukFrontendVersion` and `hmrcFrontendVersion` in
 [LibDependencies](project/LibDependencies.scala)
 
-## [6.5.0] - 2023-02-06
+## [6.6.0] - 2023-02-20
+
+### Changed
+
+- Updated the `HmrcCharacterCount` to use the underlying `GovukCharacterCount`, with translations passed in
```
oscarduignan, line 15: this looks good to me, the only thing I think is that we could add a note that there might be some action required if people's tests are the old implementations class names in their selectors, or if they're constructing the html for hmrc character count themselves, but still relying on our js, since the js that powered the previous version has been removed - can only see [1 example of this from a github search](https://github.com/hmrc/maintain-a-trust-frontend/blob/275a9f22cfaa951bb8fcb450344b17cc92b92d30/app/views/components/InputTextCharacterCount.scala.html#L59), but can see [some acceptance tests](https://github.com/hmrc/voac-acceptance-tests/blob/44361849a8aac222e1cd969d256e5abc2f07030a/src/test/resources/features/content/04_tellUsMoreAboutPermChangesToLocalArea_Content.feature#L125) using hmrc character count classes in their selectors (I don't think it's a breaking change still for play-frontend-hmrc, if we keep breaking change to mean the API is different for using our components is different, and the output can change between minor versions even if tests are using it)

JoPintoPaul: @oscarduignan Yes that's a good point. We could point this out via `CHANGELOG.md`, what do you think?

JoPintoPaul: @oscarduignan Changelog is now updated, and I'll highlight this change in internal comms channels.

oscarduignan (approved):

PR #228, opened 2023-03-27
PLATUI-2264: Validation of govukradiobutton component params reported as firing incorrectly

kyle-bowden:


src/test/scala/uk/gov/hmrc/govukfrontend/views/implicits/RichRadiosSpec.scala
```diff
@@ -63,6 +64,28 @@ class RichRadiosSpec extends AnyWordSpec with Matchers with MessagesHelpers with
...
+      val radioItem2       = RadioItem(content = Text("This is THE WORST"), value = Some("worst"), checked = false)
+
+      val radios = Radios(items = Seq(radioItem1, radioItemDivider, radioItem2)).withFormField(field)
+      radios.items shouldEqual Seq(
+        radioItem1,
+        radioItemDivider,
+        radioItem2
+      )
+    }
+
```
oscarduignan, line 88: what do you think about slightly trimming down the test? (haven't checked if this worked)

```suggestion
    "divider should not be marked as checked when field value is set to None" in {
      val emptyFormField = Field(
        form = Form(),
        name = "field",
        constraints = Nil,
        format = None,
        errors = Seq(),
        value = None
      )

      val dividerFromRadiosUpdatedWithEmptyFormField = Radios(
          items = Seq(
             RadioItem(divider = Some("or"))
          )
      ).withFormField(emptyFormField).items.head
      
      dividerFromRadiosUpdatedWithEmptyFormField shouldEqual RadioItem(
          divider = Some("or"),
          checked = false
      )
    
    }

```

kyle-bowden: @oscarduignan  Looks much better, thanks for the suggestion, i have updated the code.

PR #240, opened 2023-06-12
PLATUI-2371 restore old behaviour of `HmrcLayout` when passed `mainContentLayout = None`

ellamdav:


src/main/twirl/uk/gov/hmrc/hmrcfrontend/views/helpers/HmrcLayout.scala.html
```diff
@@ -53,6 +53,8 @@
 
 @require(!(backLinkUrl.isDefined && backLink.isDefined), "Use backLink or backLinkUrl, but not both")
 
+@noLayoutWrapper = @{ identity[Html] _ }
```
ellamdav, line 56: given this a name for clarity, but if we think the test shows intent enough then could just inline `identity`

src/test/scala/uk/gov/hmrc/hmrcfrontend/views/helpers/hmrcLayoutSpec.scala
```diff
@@ -321,6 +322,18 @@ class hmrcLayoutSpec extends AnyWordSpecLike with Matchers with JsoupHelpers wit
...
+    "not use TwoThirdsMainContent layout if passed main content layout is None" in {
+      val hmrcLayout = app.injector.instanceOf[HmrcLayout]
+      val messages   = getMessages()
+      val content    = contentAsString(
+        hmrcLayout(mainContentLayout = None)(Html("Some main content"))(fakeRequest, messages)
+      )
+      val document   = Jsoup.parse(content)
+
+      val mainContent = document.getElementById("main-content")
+      Try(s"Found unexpected content wrapper ${mainContent.child(0).outerHtml()}").toOption shouldBe None
```
ellamdav, line 334: `Try`ing to provide meaningful context if it fails, but not throw if all is well

src/main/twirl/uk/gov/hmrc/hmrcfrontend/views/helpers/HmrcLayout.scala.html
oscarduignan: I like it having a name 🙌

src/main/twirl/uk/gov/hmrc/hmrcfrontend/views/helpers/HmrcLayout.scala.html
```diff
@@ -70,7 +72,7 @@
       additionalHeadBlock = additionalHeadBlock,
       additionalScriptsBlock = additionalScriptsBlock,
       beforeContentBlock = beforeContentBlock,
-      mainContentLayout = mainContentLayout,
+      mainContentLayout = mainContentLayout.orElse(Some(noLayoutWrapper)),
       pageLayout = pageLayout,
```
oscarduignan, line 76: I think [pageLayout has a similar change in semantics](https://github.com/hmrc/play-frontend-hmrc/blob/main/src/main/twirl/uk/gov/hmrc/govukfrontend/views/components/GovukTemplate.scala.html#L90) so we might want to change that too

oscarduignan:
needs a changelog entry too I guess, I kind of guess keep it as not a major version bump since we're going back to old behaviour 🤷‍♂️

src/main/twirl/uk/gov/hmrc/hmrcfrontend/views/helpers/HmrcLayout.scala.html
ellamdav: ah, good spot, will update

ellamdav: have updated to restore the previous behaviour with `pageLayout`, but because this function is `PageLayout => Html` it's ended up tightly coupled to the implementation of `GovukTemplate` - not sure if there's a better way to do this.

oscarduignan: can't think of anything, I think in practice it won't change much so low cost I reckon

CHANGELOG.md
```diff
@@ -8,6 +8,18 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 For compatibility information see `govukFrontendVersion` and `hmrcFrontendVersion` in
 [LibDependencies](project/LibDependencies.scala)
 
+## [7.11.0] - 2023-06-13
+
+### Changed
+
+- Restore previous behaviour of `mainContentLayout` and `pageLayout` in
```
oscarduignan, line 15: @ellamdav would it be ok to say something like "in a previous version {version} the semantics of passing None changed unintentionally from causing no layout to be applied, to falling back to applying the default layouts. If you pass None to either of these parameters, please double check your usage after upgrading to this version to ensure there are no unexpected layout changes as a result."

oscarduignan (approved):

PR #254, opened 2023-10-05
PLATUI-2526 add wrapper around deprecated HmrcLayout to surface deprecation warning

ellamdav:


src/main/scala/uk/gov/hmrc/hmrcfrontend/views/html/helpers/HmrcLayout.scala
```diff
@@ -0,0 +1,77 @@
...
+import javax.inject.Inject
+
+class HmrcLayout @Inject() (
+  deprecatedHmrcLayout: DeprecatedHmrcLayout,
+  twoThirdsMainContent: TwoThirdsMainContent,
+  fixedWidthPageLayout: FixedWidthPageLayout
+) {
+  @deprecated(
+    "Use HmrcStandardPage template instead - see DeprecatedHmrcLayout for details of how HmrcLayout parameters map to HmrcStandardPage",
+    "6.3.0"
```
ellamdav, line 35: is it OK to leave this as the original version this was deprecated in?  might surprise teams on a later version who upgrade to this and start getting warnings

oscarduignan: I think it's ok to leave it - it's deprecated but there's schedule for when we're going to remove it

oscarduignan:
out of curiosity, why was it being swallowed by play?

oscarduignan (approved):

ellamdav:
> out of curiosity, why was it being swallowed by play?

Adding an annotation to the `@this()` constructor or the `@()` apply method in a Twirl template breaks compilation - the only place that compiles is against a parameter, but these don't appear to get surfaced at all.

PR #257, opened 2023-10-09
Plat UI 2515 date validation poc

ellamdav:
There are a couple of implementations of the `govUkDate` mapping - the alternate one at the bottom was our original attempt to implement using standard Play field mappings, but we found it wasn't possible to wire up the 'all/multiple fields empty' validation to fire first.

In the implementation at the top, we just bind the fields as text and then match on the shape of the input.  If we have input in all the fields, we then go on to do all the field-level validation, and return just the first error for each field.  If field-level validation passes, we then validate that it's a real date (not something like 30 Feb 2020) before returning a GovUkDate.

The GovUkDate just lets us hold onto the potentially non-numeric month that the user input for later playback - the expectation is that teams would then pass this on to their backend services as a `LocalDate`, since the actual input has already been implicitly audited on the way into the frontend.


src/main/scala/uk/gov/hmrc/govukfrontend/views/viewmodels/dateinput/DateValidationSupport.scala
```diff
@@ -0,0 +1,219 @@
...
+            }
+        }
+
+      def unbind(key: String, ymd: (String, MonthEntered, String)): Map[String, String] =
+        Map(s"$key.year" -> ymd._1, s"$key.month" -> ymd._2.entered, s"$key.day" -> ymd._3)
+    })
+
+  // this version doesn't include the GOVUK-recommended error for when all 3 fields are empty
+  // and it might be difficult/impossible to add some GOVUK-recommended errors, eg. for multiple missing fields
+  // but it uses more idiomatic Play verification, which may make it more maintainable
```
oscarduignan, line 155: not sure if a mapping is more or less idiomatic than a formatter, just that a formatter is a lower level building block

src/test/scala/uk/gov/hmrc/govukfrontend/views/viewmodels/dateinput/DateValidationSupportSpec.scala
```diff
@@ -0,0 +1,306 @@
...
+          Some(SomeFormWithDate(GovUkDate(DateEntered("1", MonthEntered("12", 12), "2023"), LocalDate.of(2023, 12, 1))))
+        )
+      }
+    }
+
+    "day is invalid or missing" should {
+      "reject invalid date where day < 1" in new Setup {
+        val form: Form[SomeFormWithDate] = testForm.bind(formData("0", "12", "2023"))
+        form.errors should be(
+          List(FormError("dateOfBirth.day", Seq("govuk.dateInput.error.date.invalid"), Seq("Date")))
```
oscarduignan, line 138: I kind of wonder - stepping back, if something useful from us - rather than an implementation, could be a table of test data that shows what error would be expected on which field - so "how to interpret and verify that you've met the current govuk guidance" that could then lead on (possibly) to a formatter that supports it - but maybe to calculate how that should look we'd want some input from service devs

I think I favour at the moment if we provide an example implementation, making it as simple as possible - so for just have a Formatter[LocalDate] and not (yet) doing extra stuff like capturing the entered values

It seems like people could still capture the month entered themselves if they did want to run away with it, for example

```scala
val whatIsYourDateOfBirthForm = Form(tuple(
  "dateOfBirth"       -> govukDate.verifying("error.youAreTooOld", _.isAfter(theMillennium)),
  "dateOfBirth.month" -> text
))

// user has gone back to edit something
val formFilledFromSomeDataStore = whatIsYourDateOfBirthForm.fill((
  LocalDate.of(2023, 10, 19),
  "Oct"
))

formFilledFromSomeDataStore("dateOfBirth.month").value // => Some("Oct")
```

and the reason being, that I think a `Formatter[LocalDate]` might not be exactly what we want in the end, but it feels like the jump from something that validates and parses a date implementing the govuk conventions to something that's optimally ergonomic for hmrc service devs feels like it would be better to do with input from service devs (which they might turn around and say _"no thanks, I will use your example as a reference of how to parse date, but I'm going to keep using my own implementation / I will just copy your one and then modify it in my service rather than use yours directly"_)

oscarduignan:
closing this draft because it was somewhat superseded by date stuff done in scaffolding https://github.com/hmrc/hmrc-frontend-scaffold.g8/pull/120

PR #259, opened 2023-10-27
Plat UI 2566 translate default back link

JoPintoPaul:


src/it/scala/uk/gov/hmrc/govukfrontend/views/components/GovukBackLinkIntegrationSpec.scala
```diff
@@ -17,16 +17,39 @@
...
+    with MessagesSupport {
+
+  protected val libraryName: String = "govuk"
+
+  protected val libraryVersion: String = govukFrontendVersion
+
+  private val component = app.injector.instanceOf[GovukBackLink]
+
+  override def render(notificationBanner: BackLink): Try[HtmlFormat.Appendable] = {
+    implicit val request: RequestHeader = FakeRequest("GET", "/foo")
```
oscarduignan, line 50: I think we might not need this implicit request since we have MessagesSupport trait mixed in - and if we remove it then we could change the param name too from notificationBanner

src/test/scala/uk/gov/hmrc/govukfrontend/views/components/GovukBackLinkSpec.scala
```diff
@@ -17,9 +17,23 @@
...
 
-class GovukBackLinkSpec extends TemplateUnitSpec[BackLink, GovukBackLink]("govukBackLink") {
+import scala.util.Try
+
+class GovukBackLinkSpec extends TemplateUnitBaseSpec[BackLink]("govukBackLink") with MessagesSupport {
+
+  private val component = app.injector.instanceOf[GovukBackLink]
+
+  def render(templateParams: BackLink): Try[HtmlFormat.Appendable] = {
+    implicit val request: RequestHeader = FakeRequest("GET", "/foo")
```
oscarduignan, line 33: same as in the integration spec, think the MessagesSupport means we don't need this implicit request

src/test/scala/uk/gov/hmrc/hmrcfrontend/views/helpers/hmrcStandardPageSpec.scala
```diff
@@ -67,26 +72,6 @@ class hmrcStandardPageSpec extends AnyWordSpecLike with Matchers with JsoupHelpe
...
-    )
-  )
-
-  private def getMessages(lang: Lang = Lang("en")): Messages = {
-    val messagesApi = stubMessagesApi(messages, new DefaultLangs(Seq(lang)))
-    new MessagesRequest(FakeRequest(), messagesApi).messages
-  }
-
-  implicit val defaultMessages: Messages = getMessages()
-
```
oscarduignan, line 89: interesting that we had some tests checking for the fallback text of the message key and some with the messages set - makes sense to me to just use our messages - easy to update tests if messages change

src/it/scala/uk/gov/hmrc/govukfrontend/views/components/GovukBackLinkIntegrationSpec.scala
```diff
@@ -17,16 +17,39 @@
 package uk.gov.hmrc.govukfrontend.views
 package components
 
+import play.api.mvc.RequestHeader
+import play.api.test.FakeRequest
+import play.twirl.api.HtmlFormat
 import uk.gov.hmrc.support.ScalaCheckUtils.ClassifyParams
 import uk.gov.hmrc.govukfrontend.support.TemplateIntegrationSpec
```
oscarduignan, line 24: think some of these imports aren't being used any more

oscarduignan:
I guess at some point we need to clarify what our versioning strategy is - are we declaring bankruptcy on meeting the versioning rules - since this would be a breaking change in that it will create a compilation error if people don't have messages in scope and that's a breaking change according to the TRG docs on library versioning. But then it's not a difficult thing to fix, and most people will probably already have it in scope, so we don't want people to avoid the upgrade because they think it's a big change 🤷 but we might be making up a problem there - I don't feel strongly either way I think

oscarduignan (approved):

PR #263, opened 2023-11-06
BDOG-2803 Build for Play 2.9

colin-lamed:
Builds for Play 2.9

This is a draft until https://github.com/hmrc/play-language/pull/47 is built.

Note, it drops usage of the deprecated `sbt-play-cross-compilation` plugin. This means that the artefact name has changed from `play-frontend-hmrc:x.y.z-play-28` to `play-frontend-hmrc-play-28:x.y.z`. For this reason it is a major version bump.

The library isn't a play service and only requires twirl and route compilation, so the use of `PlayScala` has been replaced with `SbtTwirl` and `RoutesCompiler` which cleans up the dependency tree. Integration tests still require `PlayScala` since they test via a play service.

All the twirl source locationss have been removed from the generated Build-Info (unused)

Integration tests have moved to a sub-project since `it` scope has been deprecated in sbt 1.9
Common source has moved to `src-common` as part of removing `sbt-play-cross-compilation`.


project/LibDependencies.scala
```diff
@@ -1,47 +1,43 @@
...
-import PlayCrossCompilation.dependencies
-import play.core.PlayVersion
-import play.sbt.PlayImport.ws
-import sbt.ModuleID
 
 object LibDependencies {
   val govukFrontendVersion: String = "4.7.0"
   val hmrcFrontendVersion: String  = "5.55.0"
-  val playLanguageVersion: String  = "6.2.0"
+  val playLanguageVersion: String  = "7.0.0-SNAPSHOT"
```
peteanning, line 6: Should this be a SNAPSHOT ?

colin-lamed:
Replaced by #270

PR #264, opened 2023-11-06
PLATUI-2638 enable Welsh support in accessible-autocomplete

ellamdav:
`asAccessibleAutocomplete` now requires implicit `Messages`, which could be classed as breaking, but all [current uses](https://github.com/search?q=org%3Ahmrc%20asAccessibleAutocomplete&type=code) I've spot-checked already have one in scope


TimothyFothergill (approved):
lgtm

PR #278, opened 2024-01-25
PLATUI-2706 WIP GOVUK v5 uplift

ellamdav:
* bumped govuk/hmrc-frontend versions and regenerated test fixtures


play-frontend-hmrc-play-30/src/main/scala/uk/gov/hmrc/supportfrontend/views/UtilsSupport.scala
```diff
@@ -53,6 +53,12 @@ trait UtilsSupport {
       if (s != null && s.nonEmpty) Some(s) else None
   }
 
+  def nonEmptyStringOrDefault(optValue: Option[String], default: String): String =
+    optValue match {
+      case Some(NonEmptyString(value)) => value
+      case _                           => default
+    }
+
```
oscarduignan, line 61: total stylistic thing, but what about adding some apply methods and I guess maybe an unapply for Option[String] on NonEmptyString for an api like:
```
NonEmptyString("string").getOrElse("default")
NonEmptyString(Some("string")).getOrElse("default")
```

play-frontend-hmrc-play-30/src/test/scala/uk/gov/hmrc/hmrcfrontend/views/helpers/HmrcStandardPageSpec.scala
```diff
@@ -449,7 +449,7 @@ class hmrcStandardPageSpec
 
       val exitThisPageButton = document.select(".govuk-exit-this-page__button")
       exitThisPageButton                should have size 1
-      exitThisPageButton.text()       shouldBe "Exit this page"
+      exitThisPageButton.text()       shouldBe "EmergencyExit this page"
```
oscarduignan, line 452: might not be an issue - but should there be a space here? Seems like if the emergency is visually hidden prefix then it should have a space in between. Adding comment without checking because github is crashing my browser at the moment (got it locally checked out too but don't want to have to scroll back and find this spot if it is something to check)

play-frontend-hmrc-play-30/src/test/scala/uk/gov/hmrc/hmrcfrontend/views/helpers/hmrcInternalHeadSpec.scala
```diff
@@ -56,15 +56,13 @@ class hmrcInternalHeadSpec
...
+    "include a nonce in the link tag if supplied" in {
       val hmrcInternalHead = app.injector.instanceOf[HmrcInternalHead]
       hmrcfrontend.RoutesPrefix.setPrefix("/some-service/hmrc-frontend")
       val content          = contentAsString(hmrcInternalHead()(requestWithNonce, implicitly))
 
       content should include regex
-        """<!--\[if gt IE 8\]><!-->
-          |<link href="/some-service/hmrc-frontend/assets/hmrc-frontend-\d+.\d+.\d+.min.css" nonce="a-nonce" media="all" rel="stylesheet" type="text/css" />
-          |<!--<!\[endif\]-->""".stripMargin.r
+        """<link href="/some-service/hmrc-frontend/assets/hmrc-frontend-\d+.\d+.\d+.min.css" nonce="a-nonce" media="all" rel="stylesheet" type="text/css" />""".r
```
oscarduignan, line 65: total aside, but I think maybe we don't need a nonce on css being included by links but just ones included by <style> tags inline - I'm not certain - just sharing a thought I've had in back of my mind since spending more time iwth CSP stuff

play-frontend-hmrc-play-30/src/test/scala/uk/gov/hmrc/hmrcfrontend/views/helpers/HmrcStandardPageSpec.scala
oscarduignan: looking at a fixture it has a space, so I think we need a space adding to the exit this page after the visually hidden span component

play-frontend-hmrc-play-30/src/main/scala/uk/gov/hmrc/supportfrontend/views/UtilsSupport.scala
oscarduignan: hmm , I edited my comment but it looks like it didn't save-  I think I changed my mind as I was reading - the function approach is easier to read I think, rather than apply/unapply

play-frontend-hmrc-play-30/src/test/scala/uk/gov/hmrc/hmrcfrontend/views/helpers/HmrcStandardPageSpec.scala
ellamdav: updated now

play-frontend-hmrc-play-30/src/test/scala/uk/gov/hmrc/hmrcfrontend/views/helpers/hmrcInternalHeadSpec.scala
ellamdav: Looks like `style-src` [applies to all CSS](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/style-src), whether from a `<style>` or a `<link>` tag

oscarduignan: ah, cool - so we don't have style src in the default bootstrap csp but we should keep adding it in library in case some teams do 🙌

PR #280, opened 2024-02-12
PLATUI-2844: Release the new crown logo in play-frontend-hmrc for govuk-frontend v4.8.0

kyle-bowden:


play-frontend-hmrc-play-30/src/main/scala/uk/gov/hmrc/hmrcfrontend/config/TudorCrownConfig.scala
```diff
@@ -0,0 +1,26 @@
...
+ * limitations under the License.
+ */
+
+package uk.gov.hmrc.hmrcfrontend.config
+
+import play.api.Configuration
+
+import javax.inject.Inject
+
+case class TudorCrownConfig @Inject() (config: Configuration) {
```
oscarduignan, line 23: looks fine to me, guess maybe we could make it a @Singleton so there's just one instance unless I've missed anything around that

future thing maybe- noticed that we're doing things a bit different to bootstrap - where they have their case class use simple values then they have a factory to build it from config and in a module they build it to a single instance - I guess for easier unit testing (just do `HmrcBanner(..., TudorCrownConfig(useTudorCrown=false))`) but all our stuff is doing it that way so :shrug:

https://github.com/hmrc/bootstrap-play/blob/05e34001e2b81ee4f3f851caf53e43b5bd1218d0/bootstrap-common-play-30/src/main/scala/uk/gov/hmrc/play/bootstrap/BootstrapModule.scala#L30

play-frontend-hmrc-play-30/src/main/twirl/uk/gov/hmrc/hmrcfrontend/views/components/HmrcHeader.scala.html
```diff
@@ -42,26 +47,44 @@
...
+                ></path>
+            </svg>
+            <!--[if IE 8]>
+                <img src="@assetsPath/govuk-logotype-crown.png" class="govuk-header__logotype-crown-fallback-image" width="36" height="32" alt="">
+            <![endif]-->
+        }
+        <span class="govuk-header__logotype-text">
+            GOV.UK
```
oscarduignan, line 86: gosh, not anything to change but realised this was what was throwing me earlier - it's because I've mentally moved over to expecting to see things like they would be in v5 so seeing this span with GOV.UK once was making my break think there wasn't an else - that there was just an "if use tudor crown { }" 🤦

oscarduignan:
looks good to me 😵 , just needs scalafmt and changelog entry I think

play-frontend-hmrc-play-30/src/main/scala/uk/gov/hmrc/hmrcfrontend/config/TudorCrownConfig.scala
kyle-bowden: We cant inject the configuration parameter on an object

oscarduignan (dismissed):

play-frontend-hmrc-play-30/src/main/twirl/uk/gov/hmrc/govukfrontend/views/components/GovukHeader.scala.html
```diff
@@ -49,14 +49,15 @@
...
-          ></path>
+          >
```
ellamdav, line 52: this was just breaking the line here to fix a whitespace diff

play-frontend-hmrc-play-30/src/main/twirl/uk/gov/hmrc/govukfrontend/views/components/GovukHeader.scala.html
```diff
@@ -49,14 +49,15 @@
...
+            ></path>
           </svg>
           <!--<![endif]-->
           <!--[if IE 8]>
-            <img src="@calculateAssetPath(assetsPath, "images/govuk-logotype-tudor-crown.png")" class="govuk-header__logotype-crown-fallback-image" width="36" height="32" alt="">
+            <img src="@calculateAssetPath(assetsPath, "images/govuk-logotype-tudor-crown.png")" class="govuk-header__logotype-crown-fallback-image" width="32" height="30" alt="">
```
ellamdav, line 60: fixed dimensions of fallback image

oscarduignan:
oof, good catch with govuk fixtures needing updating - makes me want a (pre)reviewers checklist in docs/maintainers of things to remember to look at that might be missing

oscarduignan:
and while fixtures are good, it also makes me want visual regression tests 😅 even though they're so much work, just have a lot more confidence when I can see a before and after

a future "what if" might be, not even doing tests comparing between screenshots, but having say a few screenshots of pages with several bits in them (like a smoke test) and just update them each time we do a change, just so that we can somewhat use the image diff functionality in github PRs to catch any issues that might have slipped by the fixture tests

PR #282, opened 2024-03-04
Plat UI 2706 govuk frontend v5 0 0 rebase

JoPintoPaul:


ellamdav:
Remember to update the release date in the CHANGELOG once approved 😅

play-frontend-hmrc-play-30/src/main/scala/uk/gov/hmrc/hmrcfrontend/views/config/StandardPhaseBanner.scala
```diff
@@ -51,7 +51,7 @@ trait StandardPhaseBanner {
 }
 
 class StandardBetaBanner extends StandardPhaseBanner {
-  def apply(url: String)(implicit messages: Messages): PhaseBanner = apply(phase = "beta", url = url)
+  def apply(url: String)(implicit messages: Messages): PhaseBanner = apply(phase = "BETA", url = url)
```
oscarduignan, line 54: should these be Beta and Alpha rather than BETA and ALPHA - the [examples on govuk](https://design-system.service.gov.uk/components/phase-banner/) are just capitalized but not fully uppercased

play-frontend-hmrc-play-30/src/main/twirl/uk/gov/hmrc/govukfrontend/views/components/GovukCookieBanner.scala.html
```diff
@@ -15,12 +15,14 @@
...
-
 @(params: CookieBanner)
 @import params._
 
-<div class="@toClasses("govuk-cookie-banner ", classes)" data-nosnippet role="region" aria-label="@ariaLabel.getOrElse("Cookie banner")"@if(hidden){ hidden}@toAttributes(attributes)>@for(message <- messages) {
-    <div class="@toClasses("govuk-cookie-banner__message govuk-width-container", message.classes)"@message.role.mapNonEmpty {role => role="@role"} @toAttributes(message.attributes) @if(message.hidden){ hidden}>
+<div class="@toClasses("govuk-cookie-banner", classes)" data-nosnippet role="region" aria-label="@{ariaLabel match {
+    case Some(NonEmptyString(str)) => str
+    case _ => "Cookie banner"
+   }}"@if(hidden){ hidden}@toAttributes(attributes)>@for(message <- messages) {
```
oscarduignan, line 24: damn mind bending twirl `@if(hidden){ hidden}` being "output 'hidden' literally" rather than the value of the hidden variable :oof:

play-frontend-hmrc-play-30/src/main/twirl/uk/gov/hmrc/govukfrontend/views/components/GovukCookieBanner.scala.html
```diff
@@ -15,12 +15,14 @@
...
+    <div class="@toClasses("govuk-cookie-banner__message", message.classes, "govuk-width-container")"@message.role.mapNonEmpty {role => role="@role"} @toAttributes(message.attributes) @if(message.hidden){ hidden}>
```
oscarduignan, line 25: guessing fixtures would catch if it wasn't right but 
```
@message.role.mapNonEmpty {role => role="@role"}
```
seems really mind bending to me too - in my head I was expecting `{role => role="@role"}` to be scala rather than a twirl block because I read @ in this case as output variable, so you'd need to be building a string - #twirlthings 😵

oscarduignan: in my head it's difference between interpretting stuff as a twirl statement
```
@if(message.role.nonEmpty) { role="@message.role"}
```
vs inline scala
```
@{if (message.role.nonEmpty) s"""role="$message.role"""" else ""}
```
guess my intuition is getting where it's `@var` a bit wrong - want like a twirl cheat sheet / test harness with table of "this renders as that"

play-frontend-hmrc-play-30/src/main/twirl/uk/gov/hmrc/govukfrontend/views/components/GovukCookieBanner.scala.html
```diff
@@ -31,36 +33,28 @@ <h2 class="govuk-cookie-banner__heading govuk-heading-m">
...
-                       content = Text(action.text),
-                       inputType = Some("button"),
-                       classes = action.classes,
-                       attributes = action.attributes,
-                       href = action.href
-                   ))
-               } else {
-                   <a class="@toClasses("govuk-link", action.classes)" href="@{action.href}"@toAttributes(action.attributes)>@action.text</a>
-               }
+            @if(action.href.getOrElse("").isEmpty || action.inputType.exists(_ == "button"))  {
```
oscarduignan, line 42: small thing to see if I missed something but why did we change this to action.href.getOrElse("").isEmpty rather than action.href.exists(_.nonEmpty) seems like more steps where exists would short circuit straight away on None

play-frontend-hmrc-play-30/src/test/scala/uk/gov/hmrc/govukfrontend/views/components/GovukPhaseBannerSpec.scala
```diff
@@ -20,7 +20,7 @@ package components
 import uk.gov.hmrc.govukfrontend.views.html.components._
 import scala.collection.JavaConverters._
 
-class GovukPhaseBannerBannerSpec extends TemplateUnitSpec[PhaseBanner, GovukPhaseBanner]("govukPhaseBanner") {
+class GovukPhaseBannerSpec extends TemplateUnitSpec[PhaseBanner, GovukPhaseBanner]("govukPhaseBanner") {
```
oscarduignan, line 23: :chef-kiss:

play-frontend-hmrc-play-30/src/test/scala/uk/gov/hmrc/govukfrontend/views/viewmodels/tasklist/TaskListItemSpec.scala
oscarduignan, line 1: Should we just have a TaskListSpec - seems like these just check if serializing works - and since these are all used in the TaskList case class I guess just running tests on that covers all these too - can see other ones have it separately too though - so guess it doesn't really matter 🤷

play-frontend-hmrc-play-30/src/test/scala/uk/gov/hmrc/hmrcfrontend/views/components/HmrcFooterSpec.scala
```diff
@@ -17,6 +17,25 @@
...
+
+class HmrcFooterSpec extends TemplateUnitBaseSpec[Footer]("hmrcFooter") with MessagesSupport {
+
+  private val component = app.injector.instanceOf[HmrcFooter]
+
+  def render(templateParams: Footer): Try[HtmlFormat.Appendable] = {
+    val request = FakeRequest("GET", "/foo")
+      .withTransientLang(templateParams.language.code)
+
+    implicit val messages: Messages = new MessagesRequest(request, messagesApi).messages
```
oscarduignan, line 37: small one, but I think we could reduce this to
```suggestion
    implicit val messages = messagesApi.preferred(Seq(Lang(templateParams.language.code)))
```
MessagesSupport trait could also be changed to `messages = mesagesApi.preferred(FakeRequest())` or use the seq in the defaultLangs in that trait rather than a fake request, but 🤷‍♂️

play-frontend-hmrc-play-30/src/test/scala/uk/gov/hmrc/hmrcfrontend/views/helpers/hmrcInternalHeadSpec.scala
```diff
@@ -73,21 +71,7 @@ class hmrcInternalHeadSpec
...
-    }
-
-    "include the hmrc-frontend minified ie8 css bundle and minified html5shiv bundle" in {
-      val hmrcInternalHead = app.injector.instanceOf[HmrcInternalHead]
-      hmrcfrontend.RoutesPrefix.setPrefix("/some-service/hmrc-frontend")
-      val content          = contentAsString(hmrcInternalHead())
-
-      content should include regex
-        """<!--\[if lte IE 8\]>
-          |<script src="/some-service/hmrc-frontend/assets/vendor/html5shiv.min.js"></script>
```
oscarduignan, line 88: Ah, just spotted that html5shiv is gone - I think we were going to keep this for a bit longer because without it the pta/bta/sca account menu breaks in IE8 - think some discussion around it relating to PLATUI-2684 in slack

play-frontend-hmrc-play-30/src/main/scala/uk/gov/hmrc/hmrcfrontend/views/config/StandardPhaseBanner.scala
ellamdav: I think we were just trying to maintain parity with what's there right now - but it feels like the research that led to this change found that Sentence case was more accessible, so maybe we should change them?  Is it maybe a question for DR?

play-frontend-hmrc-play-30/src/main/twirl/uk/gov/hmrc/govukfrontend/views/components/GovukCookieBanner.scala.html
ellamdav: I think we're trying to catch where `href` is `None` or `Some("")`, which using `exists` would probably be `!action.href.exists(_.nonEmpty)` which is arguably more mind-bending 😅 
Broadly speaking though, there are many places where we're having to code defensively purely to workaround the ambiguous JS model (see `Option[Seq[...]]`) and/or our own generative tests that populate those models with illogical inputs - it could be worth looking at some more strict typing here, or `require`ing some preconditions to make these cases go away... or it could just turn out to be yak-shaving 🤷

play-frontend-hmrc-play-30/src/test/scala/uk/gov/hmrc/govukfrontend/views/viewmodels/tasklist/TaskListItemSpec.scala
ellamdav: I guess we just followed the prevailing convention here - the test of the outer class should be enough, but these tests seem pretty quick so 🤷

play-frontend-hmrc-play-30/src/main/scala/uk/gov/hmrc/hmrcfrontend/views/config/StandardPhaseBanner.scala
JoPintoPaul: Let's use sentence case rather than capitalised and put it in the release notes.

play-frontend-hmrc-play-30/src/test/scala/uk/gov/hmrc/hmrcfrontend/views/components/HmrcFooterSpec.scala
ellamdav: Nice - works a treat 🙌

play-frontend-hmrc-play-30/src/main/twirl/uk/gov/hmrc/govukfrontend/views/components/GovukCookieBanner.scala.html
oscarduignan: ah, I get you, I totally misread the changed logic of it 🤦‍♂️

play-frontend-hmrc-play-30/src/main/scala/uk/gov/hmrc/hmrcfrontend/views/config/StandardPhaseBanner.scala
oscarduignan: I asked in int-plat-design and Tim has said to use Alpha / Beta 🙌

play-frontend-hmrc-play-30/src/test/scala/uk/gov/hmrc/hmrcfrontend/views/helpers/hmrcInternalHeadSpec.scala
oscarduignan: (this was because we're removing it and will tell maintainers of sca-wrapper and they can add it in if they want (or add a wrapping div to avoid needing extra js)

oscarduignan (approved):

oscarduignan (approved):

PR #288, opened 2024-03-21
PLATUI-2897 enable passing of attributes to input form group wrappers

ellamdav:
* add new FormGroup model
* uplift the following inputs to add attributes to govuk-form-group:
** GovukCharacterCount
** GovukCheckboxes
** GovukDateInput
** GovukFileUpload
** GovukInput
** GovukRadios
** GovukSelect
** GovukTextarea
** HmrcCharacterCount
* extract GovukFormGroupSnippet to de-duplicate repeated code across inputs
* extract GovukHintAndErrorMessageSnippet to de-duplicate existing hint/error-message snippets repeated across inputs


play-frontend-hmrc-play-30/src/main/scala/uk/gov/hmrc/hmrcfrontend/controllers/Assets.scala
```diff
@@ -27,5 +26,5 @@ import play.api.http.HttpErrorHandler
  * [RuntimeException: java.lang.NoSuchMethodError: controllers.ReverseAssets.versioned(Ljava/lang/String;)Lplay/api/mvc/Call;]
  * when using Assets.
  */
-@Singleton
+@Singleton // TODO migrate off deprecated AssetsBuilder once we drop Play 2.8 support
```
ellamdav, line 29: just to remind people who see the deprecation warning that we can't do this just yet 🙃

play-frontend-hmrc-play-30/src/test/scala/uk/gov/hmrc/govukfrontend/views/viewmodels/checkboxes/Generators.scala
```diff
@@ -63,24 +63,26 @@ object Generators {
...
-      formGroupClasses   <- genClasses()
+      formGroup          <- arbFormGroup.arbitrary
       idPrefix           <- Gen.option(genAlphaStr())
       name               <- genNonEmptyAlphaStr
       nItems             <- Gen.chooseNum(0, 5)
       items              <- Gen.listOfN(nItems, arbCheckboxItem.arbitrary)
       classes            <- genClasses()
       attributes         <- genAttributes()
-    } yield Checkboxes(
+      values             <- Gen.listOfN(nItems, genAlphaStr())
```
ellamdav, line 73: noticed `values` was missing here 🙃

play-frontend-hmrc-play-30/src/main/twirl/uk/gov/hmrc/govukfrontend/views/snippets/GovukFormGroupSnippet.scala.html
```diff
@@ -0,0 +1,25 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *@
+
+@import uk.gov.hmrc.govukfrontend.views.html.components._
+@import uk.gov.hmrc.govukfrontend.views.viewmodels.FormGroup
+
+@this()
+
+@(formGroup: FormGroup = FormGroup(), errorMessage: Option[ErrorMessage] = None)(content: => Html)
```
oscarduignan, line 22: feels a bit off passing the error message - what about a boolean and pass errorMessage.isDefined?

or considered if we could just update the FormGroup(classes=) but not sure that feels right either - nor does popping in an error flag to FormGroup case class

oscarduignan: I kind of wonder if we should just put it in components as GovukFormGroup.scala.html too - I know technically it's not a component from govuk but 🤷‍♂️

ellamdav: > feels a bit off passing the error message - what about a boolean and pass errorMessage.isDefined?
> 
> or considered if we could just update the FormGroup(classes=) but not sure that feels right either - nor does popping in an error flag to FormGroup case class

Just have a `hasErrorMessage` boolean then?  Or could even check the `content` in the snippet to see if it contains `govuk-error-message`...? 🤔  Though that may be too brittle 🤷

ellamdav: > I kind of wonder if we should just put it in components as GovukFormGroup.scala.html too - I know technically it's not a component from govuk but 🤷‍♂️

I wanted to make it clear that it's not a first-class component, but I'm not really precious about the current naming.

oscarduignan: yeah, maybe hasErrorMessage 🤷‍♂️ the more I read it the more I don't mind

we do have some other "internal" type components in the main components folder (the ones to control content layout), but then I don't know if they should really be there

I guess having snippets folder we could have a README to say "these are internal, not a public api, use at own risk as we might rejig as needed"

play-frontend-hmrc-play-30/src/main/scala/uk/gov/hmrc/govukfrontend/views/viewmodels/CommonJsonFormats.scala
```diff
@@ -22,14 +22,6 @@ import play.twirl.api.Html
...
 
 object CommonJsonFormats {
-  val readsFormGroupClasses: Reads[String] =
-    (__ \ "formGroup" \ "classes").read[String].orElse(Reads.pure(""))
-
-  val writesFormGroupClasses: OWrites[String] = new OWrites[String] {
-    override def writes(classes: String): JsObject =
-      Json.obj("formGroup" -> Json.obj("classes" -> classes))
-  }
-
```
JoPintoPaul, line 32: :meow-flamethrower:

play-frontend-hmrc-play-30/src/main/scala/uk/gov/hmrc/govukfrontend/views/viewmodels/FormGroup.scala
```diff
@@ -0,0 +1,41 @@
...
+
+import play.api.libs.functional.syntax._
+import play.api.libs.json._
+
+case class FormGroup(
+  classes: Option[String] = None,
+  attributes: Map[String, String] = Map.empty
+)
+
+object FormGroup {
```
JoPintoPaul, line 27: Suggestion: Would a FormGroup.empty method be a nice helper on the object?

play-frontend-hmrc-play-30/src/main/twirl/uk/gov/hmrc/govukfrontend/views/components/GovukCheckboxes.scala.html
```diff
@@ -17,25 +17,35 @@
...
-  @fieldset.fold(innerHtml) { fieldset =>
-    @govukFieldset(Fieldset(
-      describedBy = describedBy1.toOption,
-      legend = fieldset.legend,
-      classes = fieldset.classes,
-      attributes = fieldset.attributes,
-      html = innerHtml
-    ))
+
+@defining(fieldset.fold(describedBy)(fieldset => fieldset.describedBy.filter(_.nonEmpty).orElse(describedBy))) { describedBy =>
```
JoPintoPaul, line 29: Naming here is a bit confusing. It took me a couple of reads to work out that this is:
`@defining(fieldset.fold(params.describedBy)(fieldset => fieldset.describedBy.filter(_.nonEmpty).orElse(params.describedBy))) { fieldsetDescribedBy =>`

play-frontend-hmrc-play-30/src/main/twirl/uk/gov/hmrc/govukfrontend/views/snippets/GovukFormGroupSnippet.scala.html
JoPintoPaul: Another suggestion: rather than `snippets` as folder / package, how about `helpers`? Then we're not coupling the package too tightly for future useful things, plus matches the `hmrcfrontend` naming convention.

JoPintoPaul:
Out of interest, I notice a pattern of using `fold` on options, passing through the default as the first param list - was that just a matter of taste or is there nuance I'm missing there?

JoPintoPaul:
Also, one more comment from me - as a thought on using snippets to compose HTML templates, one thing to consider is that manually syncing with `govuk-frontend` templates will therefore require switching between files to compare. Obviously that's balanced against repetition / DRY, I just wanted to point out the risks with this divergence.

ellamdav:
> Out of interest, I notice a pattern of using `fold` on options, passing through the default as the first param list - was that just a matter of taste or is there nuance I'm missing there?

TBH I just went with what was already there, but could change to `.map(...).getOrElse(...)` if we think that's easier to grok?

play-frontend-hmrc-play-30/src/main/twirl/uk/gov/hmrc/govukfrontend/views/components/GovukCheckboxes.scala.html
ellamdav: Agree this is confusing - unusually this component allows `describedBy` to be provided in 2 ways, which complicates things.  Will rename/qualify things to make it more explicit.

play-frontend-hmrc-play-30/src/main/scala/uk/gov/hmrc/govukfrontend/views/viewmodels/FormGroup.scala
ellamdav: Agreed - I think that's more explicit 🙌

play-frontend-hmrc-play-30/src/main/twirl/uk/gov/hmrc/govukfrontend/views/components/GovukCheckboxes.scala.html
oscarduignan, line 1: some of the composition of stuff in twirl components is so hard to follow 😵 

this isn't me saying do this, just me taking a moment to think if I could imagine the a flavour I might like, this kind of structure was my first thought

```
@(checkboxes: Checkbox)

@govukFormGroup(checkboxes.formGroup) {
  @govukFieldset(checkboxes.fieldset) { 
    @govukHint(checkboxes.hint)
    @govukErrorMessage(checkboxes.errorMessage)
    ... inputs
  }
})
```

oscarduignan (dismissed):

play-frontend-hmrc-play-30/src/main/twirl/uk/gov/hmrc/govukfrontend/views/components/GovukCheckboxes.scala.html
ellamdav: Agreed - the issue seems to be accumulating the right `aria-describedby` when there's maybe a supplied one, maybe one related to the hint, and maybe one related to the error message...  But perhaps there's a nicer way to do it, like using cats `State` or something 😅

oscarduignan (approved):

PR #290, opened 2024-03-22
NOJIRA add default entry for optimizely config

oscarduignan:
to stop service getting config warnings when they set their optimizely.projectId only in app-config

### DIFF
--- a/play-frontend-hmrc-play-30/src/main/resources/reference.conf
+++ b/play-frontend-hmrc-play-30/src/main/resources/reference.conf
@@ -81,3 +81,7 @@ pta-account-menu {
 }
 
 language.fallback.url = "https://www.gov.uk/government/organisations/hm-revenue-customs"
+
+# used to enable A/B testing experiments from the experimentation team
+optimizely.url = "https://cdn.optimizely.com/js/"
+optimizely.projectId = null
--- a/play-frontend-hmrc-play-30/src/test/scala/uk/gov/hmrc/hmrcfrontend/config/TrackingConsentConfigSpec.scala
+++ b/play-frontend-hmrc-play-30/src/test/scala/uk/gov/hmrc/hmrcfrontend/config/TrackingConsentConfigSpec.scala
@@ -50,9 +50,21 @@ class TrackingConsentConfigSpec extends AnyWordSpec with Matchers {
       config.optimizelyUrl should equal(None)
     }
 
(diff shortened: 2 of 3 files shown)


play-frontend-hmrc-play-30/src/main/resources/reference.conf
```diff
@@ -81,3 +81,7 @@ pta-account-menu {
 }
 
 language.fallback.url = "https://www.gov.uk/government/organisations/hm-revenue-customs"
+
+# used to enable A/B testing experiments from the experimentation team
+optimizely.url = "https://cdn.optimizely.com/js/"
```
oscarduignan, line 86: pulled from app-config-common - better to have it here, as the value in app-config-common is actually documented as being there for play-ui so could easily get removed at some point accidentally in the future

play-frontend-hmrc-play-30/src/test/scala/uk/gov/hmrc/hmrcfrontend/config/TrackingConsentConfigSpec.scala
```diff
@@ -50,9 +50,21 @@ class TrackingConsentConfigSpec extends AnyWordSpec with Matchers {
...
+        )
+      )
+      val config                            = application.injector.instanceOf[TrackingConsentConfig]
+      config.optimizelyUrl should equal(None)
+    }
+
     "return None if optimizely.url is not defined" in {
       implicit val application: Application = buildApp(
         Map(
+          "optimizely.url"       -> null,
```
oscarduignan, line 67: this value is set by default now in reference.conf so we have to zero it out here to check that it's required

play-frontend-hmrc-play-30/src/test/scala/uk/gov/hmrc/hmrcfrontend/config/TrackingConsentConfigSpec.scala
```diff
@@ -50,9 +50,21 @@ class TrackingConsentConfigSpec extends AnyWordSpec with Matchers {
       config.optimizelyUrl should equal(None)
     }
 
+    "return None if optimizely.projectId is null" in {
```
oscarduignan, line 53: sort of tested implicitly already, but wanted to add something explicit just to make sure I'm not accidentally breaking stuff 😇

kyle-bowden (approved):

PR #292, opened 2024-04-08
PLATUI-2942: Upgrade library to govuk-frontend v5.3.0

kyle-bowden:


docs/maintainers/adr/0021-exclude-password-field.md
```diff
@@ -0,0 +1,50 @@
...
+### Immediately include the password field feature
+
+* Good, because it provides immediate feature completeness.
+* Bad, because it likely adds an unused feature for most teams, increasing maintenance overhead.
+
+### Defer the inclusion of the password field feature
+
+* Good, because it aligns our resources with current needs and reduces overhead.
+* Good, because it leaves room to adapt based on future demand.
+* Bad, because teams with immediate needs must find alternative solutions.
```
oscarduignan, line 46: LGTM extra context on this is "where we think that it would not be too troublesome / intensive for people to implement the markup themselves if they need to because it's a standard input and button some extra classes and attributes"

CHANGELOG.md
```diff
@@ -8,6 +8,18 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 For compatibility information see `govukFrontendVersion` and `hmrcFrontendVersion` in
 [LibDependencies](project/LibDependencies.scala)
 
+## [9.6.0] - 2024-04-10
+
+### Changed
+
+- Release new version of play-frontend-hmrc that includes govuk-frontend 5.3.0
+- Added ADR for [deferring the inclusion of the password field]((0021-defer-inclusion-of-password-field.md)) from govuk-frontend v5.3.0
```
oscarduignan, line 16: small one, this link doesn't work

play-frontend-hmrc-play-30/src/main/twirl/uk/gov/hmrc/govukfrontend/views/components/GovukCharacterCount.scala.html
```diff
@@ -15,38 +15,36 @@
...
+}
 
-  @if(maxLength.isEmpty && maxWords.isEmpty) {
-      @textareaDescriptionText match {
-          case Some(v) => { data-i18n.textarea-description.other="@{v}" }
-          case None => {}
-      }
-  }
+@dataAttributes = @{
+    val maybeDataAttributes = ListMap(
```
oscarduignan, line 35: do we still need to use a ListMap here? The attrs as string OR map[string, string] is really confusing to work around - looks ok, feels like it's going to be worth seeing if we can do a better comparison when testing against output rendered by govuk stuff, can imagine more an more excluded / flakey tests 😵

ellamdav: IIRC `ListMap` fixed some failing unit tests, and yes - it's awful 😅

oscarduignan (approved):

PR #296, opened 2024-04-15
PLATUI-2703 add caching to hmrc-frontend assets

TimothyFothergill:
# Purpose of PR
- Set caching on hmrc-frontend assets on services consuming this library
- Set max-age of cache to 60 minutes


play-frontend-hmrc-play-30/src/main/resources/reference.conf
```diff
@@ -82,6 +82,9 @@ pta-account-menu {
 
 language.fallback.url = "https://www.gov.uk/government/organisations/hm-revenue-customs"
 
+# Cache set to 60 minutes for hmrc-frontend assets
+play.assets.cache."/public/lib/hmrc-frontend/"="public, max-age=3600"
```
kyle-bowden, line 86: Is it maybe worth having a test around this, something like 

```
import org.scalatestplus.play._
import play.api.test.Helpers._
import play.api.test._

class AssetCachingSpec extends PlaySpec with OneAppPerTest {
  "Asset controller" should {

    "serve assets with correct cache-control headers" in {
      val request = FakeRequest(GET, "/assets/example.jpg") // Use the path to your asset
      val result = route(app, request).get

      status(result) mustBe OK
      contentType(result) mustBe Some("image/jpeg")
      header("Cache-Control", result) must contain ("max-age=31536000") // Check for the expected Cache-Control header value
    }

  }
}
```

kyle-bowden (approved):

kyle-bowden:
#

PR #297, opened 2024-04-19
PLATUI-2916: Add beforeInput(s) and afterInput(s) options to form groups in play-frontend

kyle-bowden:


play-frontend-hmrc-play-30/src/main/scala/uk/gov/hmrc/govukfrontend/views/viewmodels/FormGroup.scala
```diff
@@ -34,14 +36,30 @@ object FormGroup {
...
+      (__ \ "beforeInputs").readNullable[Content]
+  )(FormGroup.apply _)
+
+  def jsonWritesCheckboxes: OWrites[FormGroup] =
+    (
+      (__ \ "classes").writeNullable[String] and
+        (__ \ "attributes").write[Map[String, String]] and
+        (__ \ "afterInputs").writeNullable[Content] and
+        (__ \ "beforeInputs").writeNullable[Content]
+    )(unlift(FormGroup.unapply))
```
oscarduignan, line 64: I reckon we should make this something like jsonReadsForMultipleInputs / jsonWritesForMultipleInputs

because (although they have no examples at the moment) we should also use it for govukdate and govukradios formgroups - in case they add examples in the future

https://github.com/search?q=repo%3Aalphagov%2Fgovuk-frontend%20beforeInputs&type=code

play-frontend-hmrc-play-30/src/main/scala/uk/gov/hmrc/govukfrontend/views/viewmodels/FormGroup.scala
```diff
@@ -19,11 +19,13 @@ package uk.gov.hmrc.govukfrontend.views.viewmodels
...
 import play.api.libs.json._
 import uk.gov.hmrc.govukfrontend.views.viewmodels.content.Content
+import uk.gov.hmrc.govukfrontend.views.viewmodels.content.Content._
 
 case class FormGroup(
   classes: Option[String] = None,
   attributes: Map[String, String] = Map.empty,
-  afterInput: Option[Content] = None
+  afterInput: Option[Content] = None,
+  beforeInput: Option[Content] = None
```
ellamdav, line 28: can these be swapped around?  I know `afterInput` was added first, but it's just a bit weird them not being in their natural order.  I doubt anyone is already using `afterInput` so wouldn't expect it to break anything

oscarduignan: 😅 🎲 probably is ok

play-frontend-hmrc-play-30/src/main/scala/uk/gov/hmrc/govukfrontend/views/viewmodels/checkboxes/Checkboxes.scala
```diff
@@ -61,7 +62,7 @@ object Checkboxes {
         (__ \ "fieldset").readNullable[Fieldset] and
         (__ \ "hint").readNullable[Hint] and
         (__ \ "errorMessage").readNullable[ErrorMessage] and
-        (__ \ "formGroup").readWithDefault[FormGroup](defaultObject.formGroup) and
+        (__ \ "formGroup").readWithDefault[FormGroup](defaultObject.formGroup)(jsonReadsForMultipleInputs) and
```
oscarduignan, line 65: @kyle-bowden even though they don't have examples, we [should update radios and date inputs viewmodels](https://github.com/search?q=repo%3Aalphagov%2Fgovuk-frontend+beforeInputs+path%3A%2Fyaml%2F&type=code) too with this- and add the output to the twirl templates to [match govuk-frontend](https://github.com/alphagov/govuk-frontend/blob/main/packages/govuk-frontend/src/govuk/components/date-input/template.njk#L61-L63) and I guess we could add some extra fixtures to test these examples (if I'm not misremembering that capability)

oscarduignan: checked just in case and these are all the ones with just "beforeInput" in their output https://github.com/search?q=repo%3Aalphagov%2Fgovuk-frontend+%2FbeforeInput%5B%5Es%5D%2F+path%3A%2Fyaml%2F&type=code

ellamdav (approved):

PR #299, opened 2024-04-22
PLATUI-2910: Added lightweight ADR regarding decision to not use Mess…

JoPintoPaul:
…ages for hardcoded strings from hmrc-frontend inTwirl templates at this time


TimothyFothergill (approved):
lgtm

PR #301, opened 2024-04-24
PLATUI-2972 POC of metatest for generators

ellamdav:
**POC of generator metatest**
* minimal boilerplate test class required for each generator instance
* base class generates N instances using the generator, then checks for any invariant field values across those instances
* we can also ignore fields we know we don't care about - like the deprecated `language` in the `HmrcFooter`


ellamdav:
see https://github.com/hmrc/play-frontend-hmrc/pull/303 instead

PR #305, opened 2024-05-09
PLATUI-3023: Introduced GeneratorSpecs for each view model to ensure that all fields in the case classes are thoroughly tested

kyle-bowden:


play-frontend-hmrc-play-30/src/main/scala/uk/gov/hmrc/govukfrontend/views/viewmodels/accordion/Accordion.scala
```diff
@@ -39,18 +39,25 @@ object Accordion {
 
   def defaultObject: Accordion = Accordion()
 
+  // The purpose of this custom reader is to serve as a fallback for mapping JSON fields that have different names than the corresponding fields in the case class.
+  def customToggleOptionReads(fieldName: String, fallbackFieldName: String): Reads[Option[String]] =
```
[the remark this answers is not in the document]
ellamdav: or `readNullableWithFallback[String]` ?

ellamdav:
Looks like we've added `GeneratorSpec`s for every viewmodel 🙌  but I noticed a few HMRC components that don't have a `JsonRoundtripSpec` (eg. `errormessage`, `hint`, and a few others) - is it worth adding these at the same time?

play-frontend-hmrc-play-30/src/main/scala/uk/gov/hmrc/govukfrontend/views/viewmodels/accordion/Accordion.scala
```diff
@@ -39,18 +39,25 @@ object Accordion {
...
-      (__ \ "hideSectionAriaLabelText").readNullable[String] and
-      (__ \ "showSectionText").readNullable[String] and
-      (__ \ "showAllSectionsText").readNullable[String] and
-      (__ \ "showSectionAriaLabelText").readNullable[String] and
+      customToggleOptionReads("hideSectionText", "hideSection") and
+      customToggleOptionReads("hideAllSectionsText", "hideAllSections") and
+      customToggleOptionReads("hideSectionAriaLabelText", "hideSectionAriaLabel") and
+      customToggleOptionReads("showSectionText", "showSection") and
+      customToggleOptionReads("showAllSectionsText", "showAllSections") and
+      customToggleOptionReads("showSectionAriaLabelText", "showSectionAriaLabel") and
```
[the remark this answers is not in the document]
kyle-bowden: I like this idea, but I'm concerned it could evolve into a separate side project and may not be as straightforward as it appears. Perhaps we should add it to the tech debt board and revisit it later if we all agree it's something we want to address.

it-play-30/src/test/scala/uk/gov/hmrc/support/HtmlComparisonTest.scala
```diff
@@ -0,0 +1,45 @@
...
+import org.scalatest.flatspec.AnyFlatSpec
+import org.scalatest.matchers.should.Matchers
+import javax.xml.transform.TransformerFactory
+import javax.xml.transform.dom.DOMSource
+import javax.xml.transform.stream.StreamResult
+import java.io.StringWriter
+
+class HtmlComparisonTest extends AnyFlatSpec with Matchers {
+
+  def normalizeHtml(html: String): String = {
```
JoPintoPaul, line 28: `normalize` or `normalise`...?

play-frontend-hmrc-play-30/src/main/scala/uk/gov/hmrc/govukfrontend/views/viewmodels/accordion/Accordion.scala
```diff
@@ -54,5 +54,18 @@ object Accordion {
       (__ \ "rememberExpanded").readNullable[Boolean]
   )(Accordion.apply _)
 
-  implicit def jsonWrites: OWrites[Accordion] = Json.writes[Accordion]
+  implicit def jsonWrites: OWrites[Accordion] = (
+    (__ \ "id").write[String] and
```
JoPintoPaul, line 58: What's made us require an explicitly defined `OWrites` here, rather than the automatic one? Can't see it but interested to know.

play-frontend-hmrc-play-30/src/main/twirl/uk/gov/hmrc/govukfrontend/views/components/GovukErrorMessage.scala.html
```diff
@@ -14,6 +14,9 @@
  * limitations under the License.
  *@
 
+@import uk.gov.hmrc.hmrcfrontend.views.Utils._
+@import uk.gov.hmrc.govukfrontend.views.viewmodels.errormessage.ErrorMessage
```
JoPintoPaul, line 18: Are these just added to help IntelliJ?

kyle-bowden: Yes, as on my IDE the twirl template is mostly red and it's difficult to know whats  okay and whats not.

play-frontend-hmrc-play-30/src/main/twirl/uk/gov/hmrc/govukfrontend/views/components/GovukRadios.scala.html
```diff
@@ -61,7 +61,7 @@
         case None => {
           @defining(item.hint.exists(_.content.nonEmpty)) { hasHint =>
           @defining(s"$id-item-hint") { itemHintId =>
-          @defining((item.value.nonEmpty && item.value == value) || item.checked) { isChecked =>
+          @defining((item.value.nonEmpty && item.value.map(_.nonEmpty) && item.value == value) || item.checked) { isChecked =>
```
JoPintoPaul, line 64: I think this should this should be `item.value.exists(_.nonEmpty)`

play-frontend-hmrc-play-30/src/main/twirl/uk/gov/hmrc/govukfrontend/views/components/GovukSelect.scala.html
```diff
@@ -43,7 +43,7 @@
       @{formGroup.beforeInput.map(_.asHtml)}
       <select class="@toClasses("govuk-select", classes, errorMessage.fold("")(_ => "govuk-select--error"))" @attrs>
         @for(item <- items) {
-          @defining((value.nonEmpty && item.value == value) || item.selected) { isSelected =>
+          @defining((value.nonEmpty && value.map(_.nonEmpty) && item.value == value) || item.selected) { isSelected =>
```
JoPintoPaul, line 46: `item.value.exists(_.nonEmpty)`

play-frontend-hmrc-play-30/src/main/twirl/uk/gov/hmrc/govukfrontend/views/components/GovukSummaryList.scala.html
```diff
@@ -90,7 +91,6 @@
 @actionLink(action: ActionItem, cardTitle: Option[Content]) = {
     <a class="@toClasses("govuk-link", action.classes)" href="@action.href"@toAttributes(action.attributes)>@action.content.asHtml@visuallyHiddenText(action, cardTitle)</a>}
 
-@visuallyHiddenText(action: ActionItem, cardTitle: Option[Content]) = {@if(action.visuallyHiddenText.exists(_.nonEmpty) || cardTitle.exists(_.nonEmpty)) {<span class="govuk-visually-hidden">
-    @{action.visuallyHiddenText}@{cardTitle.filter(_.nonEmpty).map(ct => s" (${ct.asHtml})")}</span>}}
+@visuallyHiddenText(action: ActionItem, cardTitle: Option[Content]) = {@if(action.visuallyHiddenText.exists(_.nonEmpty) || cardTitle.isDefined) {<span class="govuk-visually-hidden"> @{action.visuallyHiddenText.map(_.rtrim)}@if(cardTitle.isDefined){@cardTitle.map {content => (@content.asHtml)}}</span>}}
```
JoPintoPaul, line 94: I see this has changed from `cardTitle.exists(_.nonEmpty)` to `cardTitle.isDefined` - is that correct? The first checks that the cardTitle is Some AND the string is non-empty, the new version will just check for Some and not None.

play-frontend-hmrc-play-30/src/main/twirl/uk/gov/hmrc/govukfrontend/views/components/GovukSelect.scala.html
```diff
@@ -43,7 +43,7 @@
       @{formGroup.beforeInput.map(_.asHtml)}
       <select class="@toClasses("govuk-select", classes, errorMessage.fold("")(_ => "govuk-select--error"))" @attrs>
         @for(item <- items) {
-          @defining((value.nonEmpty && item.value == value) || item.selected) { isSelected =>
+          @defining((value.nonEmpty && value.exists(_.nonEmpty) && item.value == value) || item.selected) { isSelected =>
```
oscarduignan, line 46: I think we can drop the initial value.nonEmpty and just have the exists(_.nonEmpty) I think if you .exists on an Option that is None then it will return false too

docs/maintainers/overview.md
```diff
@@ -98,6 +98,12 @@ library to ensure parity.
 images to be `/assets/images`, which does not match the structure of assets hosting within the `play-frontend-hmrc` 
 library.  
 
+#### ViewModel Tests
+
+To guarantee comprehensive test coverage for our view models, we have implemented `GeneratorSpec` for each model under:
+`src/test/scala/uk/gov/hmrc/govukfrontend/views/viewmodels/` and `src/test/scala/uk/gov/hmrc/hmrcfrontend/views/viewmodels/` an example GeneratorSpec can be seen here `src/test/scala/uk/gov/hmrc/govukfrontend/views/viewmodels/accordion/GeneratorSpec.scala`
+These tests confirm that all fields in the view models have associated generators. It is essential to create a corresponding `GeneratorSpec` for any new view models introduced in the future. 
+
```
JoPintoPaul, line 106: We should also add about new `GeneratorSpec` for a new component in here:
https://github.com/hmrc/play-frontend-hmrc/blob/main/docs/maintainers/upgrading.md#adding-new-components

play-frontend-hmrc-play-30/src/main/twirl/uk/gov/hmrc/govukfrontend/views/components/GovukSummaryList.scala.html
kyle-bowden: So following the nunjucks example and trying to keep as close to it as possible in our twirl templates. The cardTitle can be an empty object which causes that check to fail when its `cardTitle.exists(_.nonEmpty)` and therefore produces the incorrect output. We use `cardTitle.isDefined` to ensure we produce the same output as the nunjucks examples.

play-frontend-hmrc-play-30/src/main/scala/uk/gov/hmrc/govukfrontend/views/viewmodels/accordion/Accordion.scala
kyle-bowden: For the other question around OWrites, we need a custom writer here because the case class has fields without the text as a suffix, but the JSON examples from the nunjucks has the text as a suffix, so to keep it consistent and get tests to pass we always ensure its written with the text suffix

play-frontend-hmrc-play-30/src/test/scala/uk/gov/hmrc/helpers/GeneratorBaseSpec.scala
```diff
@@ -0,0 +1,61 @@
...
+    "generate varying data for all fields" in {
+      val instances       = generateInstances
+      val invariantFields = checkInvariantFieldValues(instances)
+      withClue("Found fields that aren't being randomly generated:") {
+        invariantFields shouldBe empty
+      }
+    }
+  }
+
+  // can't use LazyList because we need to support Scala 2.12
```
JoPintoPaul, line 43: Can we mark this as a `TODO` for when we drop Scala 2.12 support?

docs/maintainers/upgrading.md
```diff
@@ -145,6 +145,10 @@ To add a new component:
 - Add the template integration test: `src/it/scala/uk/gov/hmrc/(hmrc|govuk)frontend/views/components`. This should extend
 `TemplateIntegrationSpec[YourViewModel, YourComponent]("yourComponent")`. You will also need to create new Generator
 classes in `src/test/scala/uk/gov/hmrc/(hmrc|govuk)frontend/views/viewmodels`
+- Add a `GeneratorSpec` to you component folder, `src/main/twirl/uk/gov/hmrc/(hmrc|govuk)frontend/views/components/("yourComponent")` with contents:
```
JoPintoPaul, line 148: Should be "your component folder"

it-play-30/src/test/scala/uk/gov/hmrc/support/HtmlComparisonTest.scala
JoPintoPaul: @kyle-bowden That's weird, the functiona definition here is still reading as "normalize" for me?

play-frontend-hmrc-play-30/src/test/scala/uk/gov/hmrc/helpers/GeneratorBaseSpec.scala
```diff
@@ -0,0 +1,61 @@
...
+    "generate varying data for all fields" in {
+      val instances       = generateInstances
+      val invariantFields = checkInvariantFieldValues(instances)
+      withClue("Found fields that aren't being randomly generated:") {
+        invariantFields shouldBe empty
+      }
+    }
+  }
+
+  // TODO can't use LazyList because we need to support Scala 2.12
```
JoPintoPaul, line 43: Maybe just update this to say when we're planning to drop Scala 2.12 supprt (i.e. end May 2024)

play-frontend-hmrc-play-30/src/main/scala/uk/gov/hmrc/govukfrontend/views/viewmodels/accordion/Accordion.scala
ellamdav: Basically we've used different fieldnames in our [viewmodel](https://github.com/hmrc/play-frontend-hmrc/blob/5cf0fb5656ce527eae517fe422d15b053c5be8cb/play-frontend-hmrc-play-30/src/main/scala/uk/gov/hmrc/govukfrontend/views/viewmodels/accordion/Accordion.scala#L29-L34) than the names used by [govuk-frontend](https://github.com/alphagov/govuk-frontend/blob/main/packages/govuk-frontend/src/govuk/components/accordion/accordion.yaml#L22-L44), so without a custom writer to fix up the field names, the nunjucks component won't see any of those fields.

Ideally I'd like to deprecate the existing, incorrect, field names and realign them with govuk-frontend, but we said we'd kick that into tech debt for now.

play-frontend-hmrc-play-30/src/test/scala/uk/gov/hmrc/helpers/GeneratorBaseSpec.scala
ellamdav: Yeah, I'd take the `TODO` off here because we can't do it yet - but it's just to warn anyone who sees a deprecation warning and wastes time trying to fix it before we've dropped Scala 2.12

PR #309, opened 2024-06-06
PLATUI-2901: Added new methods withDayMonthYear and withDayMonthYearW…

JoPintoPaul:
…ithErrorAsHtml

### DIFF
--- a/README.md
+++ b/README.md
@@ -317,81 +317,52 @@ the content as below:
 
 #### RichDateInput
 
-The implicit class `RichDateInput` provides an extension method `withFormField(field: play.api.data.Field)`
-for the `DateInput` view model.
+The implicit class `RichDateInput` provides three extension methods to populate the `DateInput` view model:
+* `withDayMonthYearFormField(field: Field)`
+* `withDayMonthFormField(field: Field)`
+* `withMonthYearFormField(field: Field)`
 
-This method takes a Play `Field` and enriches the `DateInput` with:
-* three InputItems corresponding to the day, month and year fields with
-    * ids and names of the form `<name>.day`, `<name>.month` and `<name>.year`
-    * standard English and Welsh labels
+These methods takes a Play `play.api.data.Field` and enrich the `DateInput` with:
+* InputItems corresponding to a combination of the day, month and year fields with
+    * ids and names of the form `<name>.day`, `<name>.month` and `<name>.year`  (`<name>` being populated from the `Field` name if present, else `date`).
+* standard English and Welsh labels
 * an id, set to the `Field` name
 * values for the nested day, month and year fields
 * a Text error message from an implicit `Messages` indexed by the first non-empty:
-    * `field("day").error`
-    * `field("month").error`
-    * `field("year").error`
-    * `field.error`
-* the correct CSS error classes applied to any invalid nested day, month or year field or
-  to all fields in the case of a global date error.
-
-This helper assumes the Field passed to it consists of three [nested fields](https://www.playframework.com/documentation/2.8.x/ScalaForms#Nested-values)
-indexed by the keys `day`, `month` and `year`.
+* `field("day").error`
+* `field("month").error`
+* `field("year").error`
+* `field.error`
+* the correct CSS error classes applied to any invalid nested day, month or year field or  to all fields in the case of a global date error.
 
 For example, if using the one question per page pattern, the method could be used as follows:
 
-```scala
-@govukDateInput(DateInput(
-  hint = Some(Hint(content = Text("date.hint"))),
-  fieldset = Some(Fieldset(
-    legend = Some(Legend(
-      content = Text(messages("date.heading")),
-      classes = "govuk-fieldset__legend--l",
-      isPageHeading = true)))
-  )
-).withFormField(dateInputForm("date")))
-```
+```scala  
+@govukDateInput(DateInput(  
+ hint = Some(Hint(content = Text("date.hint"))), fieldset = Some(Fieldset( legend = Some(Legend( content = Text(messages("date.heading")), classes = "govuk-fieldset__legend--l", isPageHeading = true))) )).withDayMonthYearFormField(dateInputForm("date")))  
+```  
 
 Setting up form validation for this field might look like:
 
-```scala
-case class DateData(day: String, month: String, year: String)
-case class PageData(date: DateData)
+```scala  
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -317,81 +317,52 @@ the content as below:
...
-      content = Text(messages("date.heading")),
-      classes = "govuk-fieldset__legend--l",
-      isPageHeading = true)))
-  )
-).withFormField(dateInputForm("date")))
-```
+```scala  
+@govukDateInput(DateInput(  
+ hint = Some(Hint(content = Text("date.hint"))), fieldset = Some(Fieldset( legend = Some(Legend( content = Text(messages("date.heading")), classes = "govuk-fieldset__legend--l", isPageHeading = true))) )).withDayMonthYearFormField(dateInputForm("date")))  
+```
```
oscarduignan, line 343: I think it might be worth formatting these two code samples with line breaks again

play-frontend-hmrc-play-30/src/main/scala/uk/gov/hmrc/hmrcfrontend/views/implicits/RichDateInputSupport.scala
```diff
@@ -65,35 +145,72 @@ trait RichDateInputSupport {
...
+        value = if (inputItem.value == defaultInputItem.value) field(key).value else inputItem.value,
+        label = if (inputItem.label == defaultInputItem.label) Some(messages(s"date.input.$key")) else inputItem.label,
+        classes = s"$classes ${errorClass(field(key))}".trim
+      )
+    }
 
-      val dateInputItems = if (dateInput.items.size == 3) dateInput.items else Seq.fill(3)(InputItem.defaultObject)
+    private def defaultInputItem(field: Field, key: String, className: String): InputItem = {
+      def errorClass(itemField: Field) =
+        if (field.errors.nonEmpty || itemField.errors.nonEmpty) "govuk-input--error" else ""
```
oscarduignan, line 179: I think we've got this same bit in two places now, it could be hoisted up a level as a private def

play-frontend-hmrc-play-30/src/main/scala/uk/gov/hmrc/hmrcfrontend/views/implicits/RichDateInputSupport.scala
```diff
@@ -65,35 +145,72 @@ trait RichDateInputSupport {
...
-        inputItem(dateInputItems(0), "day", className = "govuk-input--width-2"),
-        inputItem(dateInputItems(1), "month", className = "govuk-input--width-2"),
-        inputItem(dateInputItems(2), "year", className = "govuk-input--width-4")
+      InputItem(
+        id = s"${field.name}.$key",
+        name = s"${field.name}.$key",
+        value = field(key).value,
+        label = Some(messages(s"date.input.$key")),
+        classes = s"$className ${errorClass(field(key))}".trim
       )
```
oscarduignan, line 187: we've lost the ability here to pass existing values for InputItems, and I can't remember if this was a deliberate choice - this might impact users if they need to set any custom attributes, they might not be able to just swap from `withFormField` to `withDayMonthYearInputItems`

I wonder if this is a good chance to add in some `require()` guardrails around stuff since it's new api - so for example, if we stick with throwing away anything in existing items, then we could require items be empty, if we keep the existing items then we could do some checks like - with the current one we use the inputItem.name if it exists, except we don't check if `InputItem.name == "${field.name}.$key"` so theoretically someone might not realise that errors weren't being wired up as they expected if their input items weren't in the "${field.name}.(day|month|year)" format

not 100% on it - I might be sending us on a side quest

README.md
JoPintoPaul: @oscarduignan Sure thing 🙌

play-frontend-hmrc-play-30/src/main/scala/uk/gov/hmrc/hmrcfrontend/views/implicits/RichDateInputSupport.scala
JoPintoPaul: @oscarduignan On this one, we have similar but not identical methods for `deprecatedInputItem` and `defaultInputItem`. In my head I split them out because otherwise we have to munge in some unnecessary stuff - but happy to talk this one through?

README.md
JoPintoPaul: @oscarduignan Done

play-frontend-hmrc-play-30/src/main/scala/uk/gov/hmrc/hmrcfrontend/views/implicits/RichDateInputSupport.scala
JoPintoPaul: My recall of this was that we're deliberately simplifying things so that teams can no longer pass through custom values to the `with...(field: Field)` helpers, as the idea was to make the defaults much simpler to use, then if people want to customise they can use their own. I do agree though that it is a change, and the idea of using a `@require` is a good one to make this super visible.

JoPintoPaul: @oscarduignan Added `require` to the helper methods.

play-frontend-hmrc-play-30/src/main/scala/uk/gov/hmrc/hmrcfrontend/views/implicits/RichDateInputSupport.scala
oscarduignan: that makes sense! save future accidents

play-frontend-hmrc-play-30/src/main/scala/uk/gov/hmrc/hmrcfrontend/views/implicits/RichDateInputSupport.scala
oscarduignan: 🙌

oscarduignan: looks good to me, that means that people won't get caught out when they move over

oscarduignan (approved):

PR #311, opened 2024-06-24
PLATUI-2603: make serviceUrl optional

oscarduignan:
PLATUI-2603: bump hmrc-frontend to 6.21.0


play-frontend-hmrc-play-30/src/main/twirl/uk/gov/hmrc/hmrcfrontend/views/components/HmrcInternalHeader.scala.html
```diff
@@ -25,49 +25,53 @@
...
+        @if(serviceUrl.nonEmpty) {
+          <a href="@serviceUrl" class="hmrc-internal-header__link">
+            @serviceName
+          </a>
+        } else {
+          @serviceName
+        }
+      </div>
+    }
+  </div>
```
oscarduignan, line 76: restructured this slightly because I thought if's were a bit simpler to read

play-frontend-hmrc-play-30/src/main/twirl/uk/gov/hmrc/hmrcfrontend/views/components/HmrcInternalHeader.scala.html
```diff
@@ -25,49 +25,53 @@
...
-        </a>
-      </div>}
-    case _ => {}
-  }</div>
-
+      </a>
+    </div>
+    @if(serviceName.exists(_.nonEmpty)) {
+      <div class="hmrc-internal-header__service-name">
+        @if(serviceUrl.nonEmpty) {
```
oscarduignan, line 67: didn't want to change the api to Option[String] because it would still have the same behaviour since empty string is falsey in js and the nunjucks doesn't check if it's undefined, it just checks if it's truthy/falsey

(though it also means that js could happily handle serviceUrl=false where as scala one wouldn't handle that)

play-frontend-hmrc-play-30/src/main/twirl/uk/gov/hmrc/hmrcfrontend/views/components/HmrcInternalHeader.scala.html
oscarduignan, line 1: everything else was just whitespace

kyle-bowden (approved):

PR #312, opened 2024-06-25
PLATUI-3071 - Uplift hmrc-frontend to 6.22.0

TimothyFothergill:
# Purpose of PR
- Bump `hmrc-frontend` to 6.22.0.
  - This is to pull the Welsh language equivalent to 'Account' ('Cyfrif') in `aria-label` under `account-menu`.
- Update `aria-label` in `HmrcAccountMenu` to be set based on language.


oscarduignan:
@TimothyFothergill I think we need to run `sbt generateHmrcFixtures` from https://github.com/hmrc/play-frontend-hmrc/blob/main/docs/maintainers/upgrading.md#basic-steps might not change much but there would always be a fixtures version file updated

oscarduignan (approved):

PR #316, opened 2024-07-17
PLATUI-3035 remove unused url parameter from report-technical-problem

TimothyFothergill:
# Purpose of PR
- Remove unused `newTab` url parameter from the `report-technical-problem` component.
  - No analytics could be found using this parameter, so removing it now. If it turns out we need to revert this, we can do so.


play-frontend-hmrc-play-30/src/test/scala/uk/gov/hmrc/hmrcfrontend/views/helpers/hmrcReportTechnicalIssueHelperSpec.scala
```diff
@@ -77,7 +77,7 @@ class hmrcReportTechnicalIssueHelperSpec
       links should have size 1
       links.first.attr(
         "href"
-      )   shouldBe "https://www.tax.service.gov.uk/contact/report-technical-problem?newTab=true&service=online-payments&referrerUrl=https%3A%2F%2Fwww.tax.service.gov.uk%2Ffoo"
+      )   shouldBe "https://www.tax.service.gov.uk/contact/report-technical-problem?service=online-payments&referrerUrl=https%3A%2F%2Fwww.tax.service.gov.uk%2Ffoo"
```
oscarduignan, line 80: jsoup is decoding it for us too! But I think it's ok in this case because we've got the fixture tests and the test in hmrc-frontend to ensure it's encoded correctly

oscarduignan (approved):

PR #317, opened 2024-08-05
PLATUI-3170

TimothyFothergill:
# Purpose of PR

- Bump hmrc-frontend to 6.27.0
- Bump play-frontend-hmrc to 5.4.1
  - This comes with complications. There are a number of changes in 5.4.1 such as:
    - Updating breadcrumbs to utilise `nav` elements and `aria-labels`
    - Minor tweaks to `GovukCheckboxes`, `GovukRadios` and `GovukSelect`
  - This update comes with a breaking change to the way we handle govuk back-links
    - This change will be addressed in a future release of govuk-frontend, either reverting the existing change or a new change.
    - Decided the best course of action is to move fixtures to excluded until a new patch comes in to revert back-link changes
- CHANGELOG updated

### DIFF
--- a/play-frontend-hmrc-play-30/src/main/twirl/uk/gov/hmrc/govukfrontend/views/components/GovukBackLink.scala.html
+++ b/play-frontend-hmrc-play-30/src/main/twirl/uk/gov/hmrc/govukfrontend/views/components/GovukBackLink.scala.html
@@ -18,7 +18,7 @@
 
 @(params: BackLink)(implicit messages: Messages)
 @import params._
-<a href="@{"".?:(href)}" class="@toClasses("govuk-back-link", classes)"@toAttributes(attributes)>@content match {
+<a href="@{"#".?:(href)}" class="@toClasses("govuk-back-link", classes)"@toAttributes(attributes)>@content match {
   case NonEmptyHtml(html) => {@html}
   case NonEmptyText(text) => {@text}
   case _ => {@messages("back.text")}
--- a/play-frontend-hmrc-play-30/src/test/scala/uk/gov/hmrc/govukfrontend/views/components/GovukBackLinkSpec.scala
+++ b/play-frontend-hmrc-play-30/src/test/scala/uk/gov/hmrc/govukfrontend/views/components/GovukBackLinkSpec.scala
@@ -88,5 +88,15 @@ class GovukBackLinkSpec extends TemplateUnitBaseSpec[BackLink]("govukBackLink") 
       output.attr("data-test")  shouldBe "attribute"
       output.attr("aria-label") shouldBe "Back to home"
     }
+
+    // This test is a temporary measure as in govuk-frontend 5.4.1, the way default values are handled
+    // have changed. A change to revert this has been discussed, so this test is only here until this
+    // has taken place.
+    "ensure an empty href value returns #" in {
+      val params = BackLink()
+      val output = component(params).select(".govuk-back-link")
+
+      output.attr("href") shouldBe "#"
+    }
   }
 }
--- a/play-frontend-hmrc-play-30/src/test/scala/uk/gov/hmrc/govukfrontend/views/viewmodels/backlink/Generators.scala
+++ b/play-frontend-hmrc-play-30/src/test/scala/uk/gov/hmrc/govukfrontend/views/viewmodels/backlink/Generators.scala
@@ -17,14 +17,14 @@
 package uk.gov.hmrc.govukfrontend.views.viewmodels.backlink
 
 import org.scalacheck.Arbitrary
-import uk.gov.hmrc.govukfrontend.views.viewmodels.Generators.{genAlphaStr, genAttributes, genClasses}
+import uk.gov.hmrc.govukfrontend.views.viewmodels.Generators.{genAlphaStr, genAttributes, genClasses, genNonEmptyAlphaStr}
 import uk.gov.hmrc.govukfrontend.views.viewmodels.content.Generators._
 
 object Generators {
 
   implicit val arbBackLink: Arbitrary[BackLink] = Arbitrary {
     for {
-      href       <- genAlphaStr()
+      href       <- genNonEmptyAlphaStr
       classes    <- genClasses()
       attributes <- genAttributes()
       content    <- arbContent.arbitrary


TimothyFothergill:
> * Runtime issues may occur. Change `org.playframework.sbt-plugin` to `3.0.3` in [project/plugins.sbt](https://github.com/hmrc/play-frontend-hmrc/blob/PLATUI-3170/project/plugins.sbt#L20) - to match what `org.playframework.play-filters-helpers` uses

Not bumping to 3.0.3 as Templates are no longer implementing the Template trait: https://github.com/playframework/twirl/commit/75eec7cf46e23f0d10fa0a7f24d2f9b44c6c9738

play-frontend-hmrc-play-30/src/main/scala/uk/gov/hmrc/govukfrontend/views/viewmodels/breadcrumbs/Breadcrumbs.scala
```diff
@@ -26,6 +26,7 @@ case class Breadcrumbs(
   items: Seq[BreadcrumbsItem] = Nil,
   classes: String = "",
   attributes: Map[String, String] = Map.empty,
+  labelText: Option[String] = None,
```
oscarduignan, line 29: sorry, taken me ages to actually get to reviewing

I don't think it matters much for us since we don't have to worry much about binary compatibility of our case classes, and from, a quick search, people are using param names to instantiate this (and I've definitely done it without thinking in the past)

but the best place to put new params is at the end, just in case someone is passing params positionally that way it won't break for them - and if binary compatibility was an issue then that would be the best place to put new params too I think to keep apply methods consistently ordered

play-frontend-hmrc-play-30/src/main/twirl/uk/gov/hmrc/govukfrontend/views/components/GovukCheckboxes.scala.html
```diff
@@ -61,23 +61,25 @@
...
-                attributes = item.label.map(_.attributes).getOrElse(Map.empty),
-                content = item.content
-              ))
-              @if(hasHint(item)) {
-                @govukHint(Hint(
-                  id = Some(itemHintId(item, index)),
-                  classes = toClasses("govuk-checkboxes__hint", item.hint.map(_.classes).getOrElse("")),
-                  attributes = item.hint.map(_.attributes).getOrElse(Map.empty),
-                  content = item.hint.map(_.content).getOrElse(Empty)
+          @if(item.value.nonEmpty){
```
oscarduignan, line 64: hmm, where did this change come from? I can't see a similar check in govuk-frontend 

https://github.com/alphagov/govuk-frontend/blame/main/packages/govuk-frontend/src/govuk/components/checkboxes/template.njk

play-frontend-hmrc-play-30/src/main/twirl/uk/gov/hmrc/govukfrontend/views/components/GovukSelect.scala.html
```diff
@@ -46,9 +46,9 @@
           @defining((value.exists(_.nonEmpty) && item.value == value) || item.selected) { isSelected =>
             @* Allow selecting by text content (the value for an option when no value attribute is specified) *@
             @defining(item.value.getOrElse(item.text)) { effectiveValue =>
-              <option @if(item.value.nonEmpty) {value="@item.value"}
+              @if(item.value.nonEmpty || isSelected || item.disabled || !item.attributes.isEmpty || item.text != ""){<option @if(item.value.nonEmpty) {value="@item.value"}
```
oscarduignan, line 49: same for this one, is it recreating this

https://github.com/alphagov/govuk-frontend/blob/main/packages/govuk-frontend/src/govuk/components/select/template.njk#L51

so allowing falsey items - like if someone passed `items = [ false, undefined, { .. actual item ... } ]`?

I think for these, like with the accordion we can exclude support for these cases, because people will be prevented by the type system from entering nonsense so we don't need to account for it

play-frontend-hmrc-play-30/src/test/scala/uk/gov/hmrc/govukfrontend/views/components/GovukBreadcrumbsSpec.scala
```diff
@@ -43,7 +45,9 @@ class GovukBreadcrumbsSpec extends TemplateUnitSpec[Breadcrumbs, GovukBreadcrumb
       )
       val output = component(params)
 
-      output.body should include("<div class=\"govuk-breadcrumbs\" role=\"navigation\" id=\"my-navigation\">")
+      output.body should include(
+        "<nav aria-label=\"Breadcrumb\" class=\"govuk-breadcrumbs\" role=\"navigation\" id=\"my-navigation\">"
+      )
```
oscarduignan, line 50: not sure why we have these tests really

play-frontend-hmrc-play-30/src/main/twirl/uk/gov/hmrc/govukfrontend/views/components/GovukSelect.scala.html
TimothyFothergill: Just had a dig through and tested the previous one and with this - And I think you've got it. I believe these were failing the fixtures, the changes I proposed "fixed" those issues and then when we uncovered the falsey -> falsy name change, I updated the excluded-fixtures to the correct names. 

I think these fixtures are now not being checked anyway (At least it's not failing any more with these changes removed). Will commit the removal of the changes for checkboxes and select 🗑️

play-frontend-hmrc-play-30/src/main/scala/uk/gov/hmrc/govukfrontend/views/viewmodels/backlink/BackLink.scala
```diff
@@ -34,7 +34,7 @@ object BackLink {
   def defaultObject: BackLink = BackLink()
 
   def mimicsBrowserBackButtonViaJavaScript(implicit messages: Messages): BackLink = BackLink.withDefaultText(
-    href = "#",
+    href = "",
```
oscarduignan, line 37: This should stay as #

play-frontend-hmrc-play-30/src/main/twirl/uk/gov/hmrc/govukfrontend/views/components/GovukBackLink.scala.html
```diff
@@ -18,7 +18,7 @@
 
 @(params: BackLink)(implicit messages: Messages)
 @import params._
-<a href="@{"#".?:(href)}" class="@toClasses("govuk-back-link", classes)"@toAttributes(attributes)>@content match {
+<a href="@{"".?:(href)}" class="@toClasses("govuk-back-link", classes)"@toAttributes(attributes)>@content match {
```
oscarduignan, line 21: since we've disabled the tests that compare output with govuk, I actually wonder if we should keep this the current way - since we know it is going to be reverted in govuk-frontend

play-frontend-hmrc-play-30/src/main/twirl/uk/gov/hmrc/govukfrontend/views/components/GovukSelect.scala.html
oscarduignan: sweet 🙌

play-frontend-hmrc-play-30/src/main/twirl/uk/gov/hmrc/govukfrontend/views/components/GovukBackLink.scala.html
TimothyFothergill: I think since we know we'll be going back to the "#" href, once the change comes in, we should keep it as is. I'll get this put back to include the "#".

play-frontend-hmrc-play-30/src/test/scala/uk/gov/hmrc/govukfrontend/views/components/GovukBreadcrumbsSpec.scala
TimothyFothergill: Yeah, I just updated the existing test here. I'll put a ticket up for tech debt to investigate whether we have redundant tests, as I think that could be a separate bit of work.

play-frontend-hmrc-play-30/src/main/scala/uk/gov/hmrc/govukfrontend/views/viewmodels/breadcrumbs/Breadcrumbs.scala
TimothyFothergill: Fair point, moved!

play-frontend-hmrc-play-30/src/test/scala/uk/gov/hmrc/govukfrontend/views/components/GovukBreadcrumbsSpec.scala
TimothyFothergill: Tech debt ticket raised to investigate all play-frontend-hmrc tests we think are redundant/unnecessary

oscarduignan (approved):

PR #325, opened 2024-10-02
PLATUI-3297 fix bug with service navigation twirl

TimothyFothergill:
# Purpose of PR
- To fix a bug with the `ServiceNavigation` component twirl template
  - This addresses a discrepancy between our `twirl` template and the `govuk-frontend` `nunjucks` equivalent
  - The slotted content would always provide a `nav` element if any slot was filled, but we only want the `nav` element if a `navigationStart` or `navigationEnd` slot has been filled with content
- Add a new test and update an existing test around this edge case

### DIFF
--- a/play-frontend-hmrc-play-30/src/test/scala/uk/gov/hmrc/govukfrontend/views/components/GovukServiceNavigationSpec.scala
+++ b/play-frontend-hmrc-play-30/src/test/scala/uk/gov/hmrc/govukfrontend/views/components/GovukServiceNavigationSpec.scala
@@ -41,7 +41,7 @@ class GovukServiceNavigationSpec
       output.first().text() shouldBe "my-service"
     }
 
-    "render ServiceNavigationItems correctly" in {
+    "render ServiceNavigationItems correctly in a nav element" in {
       val params = ServiceNavigation(
         navigation = Seq(
           ServiceNavigationItem(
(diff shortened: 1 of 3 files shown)


play-frontend-hmrc-play-30/src/test/scala/uk/gov/hmrc/govukfrontend/views/components/GovukServiceNavigationSpec.scala
```diff
@@ -70,7 +70,7 @@ class GovukServiceNavigationSpec
       output.first().attr("href") shouldBe "#"
     }
 
-    "render ServiceNavigation with slotted content" in {
+    "render ServiceNavigation with slotted content out of the nav element" in {
```
oscarduignan, line 73: what about renaming this something like

"render the nav element even if there are no nav items when either of the navigation slots are provided"

oscarduignan (approved):

play-frontend-hmrc-play-30/src/test/scala/uk/gov/hmrc/govukfrontend/views/components/GovukServiceNavigationSpec.scala
TimothyFothergill: This test ensures we do **not** render the nav element, as neither the navigation nor a navigation Slot is added. Should I add a third test for when we have a nav element without it being slotted content?

JoPintoPaul (approved):

PR #330, opened 2024-11-01
PLATUI-3334: add test for unbalanced html tags

oscarduignan:
And fix a bug with GovukTaskList if items were empty

### DIFF
--- a/play-frontend-hmrc-play-30/src/main/twirl/uk/gov/hmrc/govukfrontend/views/components/GovukServiceNavigation.scala.html
+++ b/play-frontend-hmrc-play-30/src/main/twirl/uk/gov/hmrc/govukfrontend/views/components/GovukServiceNavigation.scala.html
@@ -22,68 +22,69 @@
 @(params: ServiceNavigation)(implicit messages: Messages)
 @import params._
 
-@if(params.serviceName.isDefined || params.slots.flatMap(_.start).isDefined || params.slots.flatMap(_.end).isDefined) {
-  <section aria-label="@{ariaLabel.getOrElse(messages("govuk.serviceNavigation.ariaLabel"))}" class="@toClasses("govuk-service-navigation", classes)" data-module="govuk-service-navigation" @toAttributes(attributes)>
-} else {
-  <div class="@toClasses("govuk-service-navigation", classes)" data-module="govuk-service-navigation" @toAttributes(attributes)>
+@govukServiceNavigationContainer = {
+  <div class="govuk-service-navigation__container">
+    @if(serviceName.nonEmpty) {
+      <span class="govuk-service-navigation__service-name">
+        @serviceUrl match {
+          case Some(url) => {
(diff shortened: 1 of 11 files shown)


play-frontend-hmrc-play-30/src/test/resources/fixtures/govuk-frontend/additional-fixtures/template-customised/input.json
oscarduignan, line 1: it is technically valid to not close a <p> but we don't need to have tests with that, and I'd recommend against it

play-frontend-hmrc-play-30/src/main/twirl/uk/gov/hmrc/govukfrontend/views/components/GovukServiceNavigation.scala.html
```diff
@@ -22,68 +22,69 @@
...
+}
+
 @if(params.serviceName.isDefined || params.slots.flatMap(_.start).isDefined || params.slots.flatMap(_.end).isDefined) {
   <section aria-label="@{ariaLabel.getOrElse(messages("govuk.serviceNavigation.ariaLabel"))}" class="@toClasses("govuk-service-navigation", classes)" data-module="govuk-service-navigation" @toAttributes(attributes)>
+    @govukWidthContainer
+  </section>
 } else {
   <div class="@toClasses("govuk-service-navigation", classes)" data-module="govuk-service-navigation" @toAttributes(attributes)>
+    @govukWidthContainer
+  </div>
```
oscarduignan, line 89: it's easy to miss balancing tags when you have multiple if statements that need to change together, so rather than that, I've extracted some variables - now we could get help from IDE too if stuff is unbalanced (though won't pick up everything)

oscarduignan: you can see the failing test cases for this if you look at the build of the last commit, where I added tests to check for unbalanced html

oscarduignan: I wiped had to rebase on main, so I've lost the build - but here was the test result showing it failing on unbalanced tags in service navigation

CleanShot 2024-11-04 at 16 10 10@2x

play-frontend-hmrc-play-30/src/main/twirl/uk/gov/hmrc/govukfrontend/views/components/GovukTaskList.scala.html
oscarduignan, line 1: only a problem when item is empty, which seems unlikely, and is possibly us matching some semantics of js that we don't need to in scala - but in that case you'd have an unopened </li>

TimothyFothergill (approved):
lgtm

PR #333, opened 2024-11-19
PLATUI-3278 bump hmrc-frontend version

TimothyFothergill:
# Purpose of PR
- Bump `hmrc-frontend` to `6.41.0`
  - This change introduces:
    - `accessible-autocomplete` upgrade to `3.0.1` from `2.0.4`
      - This introduces some style changes, so teams should be sure to check doesn't impact your services.
- This also pulls in `hmrc-frontend` changes in `6.40.0`, which includes changes to the govuk header line-height values, setting them to unitless values and ensuring the baseline matches the GOV.UK logo and service name


oscarduignan (approved):

oscarduignan:
the-lord-of-the-rings-gandalf

PR #338, opened 2025-01-09
PLATUI-3352: update hmrc-frontend to 6.51.0

oscarduignan:
when you interact with the accessible autocomplete, it updates the aria-describedby attribute to add/remove link to it's assistiveHint

this was removing the link to the error/hint associated with the underlying select as well, this updated version of hmrc-frontend has some javascript to maintain the error/hint in the aria-describedby attribute, because the removal of them might be a wcag failure

https://github.com/user-attachments/assets/17affba2-2a60-4dad-a178-5badc6a59815


oscarduignan:
will be superseded by another PR that's updating govuk-frontend version

PR #343, opened 2025-03-13
NOJIRA consider how to enable service navigation

oscarduignan:
was just looking at ways to migrate off of current Header if we do want to add some options for serviceNavigation to it since we're basically at the 22 value limit for case classes so need to refactor

didn't want to suggest we do what we did with hmrc standard layout again - where we created a new component and delegated as that seemed like it involved work for teams that was confusing

this approach instead is changing input param and having an implicit conversion from old Header to new input param type so people using Header don't have to make any changes and we just have one template to maintain

~not sure exactly how to enable it via config at this point - some nuance to how it needs to be possible to enable it, future spikes, but wanted to create this as a prompt for that work~

added some stuff around how we could enable it via config by using null as a sentinel value (so can be overridden)

### DIFF
--- a/play-frontend-hmrc-play-30/src/main/twirl/uk/gov/hmrc/hmrcfrontend/views/helpers/HmrcStandardHeader.scala.html
+++ b/play-frontend-hmrc-play-30/src/main/twirl/uk/gov/hmrc/hmrcfrontend/views/helpers/HmrcStandardHeader.scala.html
@@ -17,6 +17,7 @@
 @import uk.gov.hmrc.hmrcfrontend.views.html.components.HmrcHeader
 @import uk.gov.hmrc.hmrcfrontend.views.Aliases.{Cy, En, Header, UserResearchBanner}
 @import uk.gov.hmrc.govukfrontend.views.Aliases.PhaseBanner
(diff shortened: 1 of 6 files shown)


play-frontend-hmrc-play-30/src/main/twirl/uk/gov/hmrc/hmrcfrontend/views/helpers/HmrcStandardHeader.scala.html
```diff
@@ -28,11 +29,16 @@
...
   displayHmrcBanner: Boolean = false,
   additionalBannersBlock: Option[Html] = None,
-  containerClasses: String = Header.defaultObject.containerClasses
+  containerClasses: String = Header.defaultObject.containerClasses,
+  serviceNavigation: Option[ServiceNavigation] = null // null used as sentinel to allow turning it off even if enabled in config
 )(implicit messages: Messages, request: RequestHeader)
+@**
+here we could check if serviceNavigation == null and if it does, grab the default from config, or else just use the
+supplied serviceNavigation - where using null isn't great but it's only used as an internal thing never exposed to users
+**@
```
oscarduignan, line 38: we'd also need serviceNavigation option like this added to layout components

PR #347, opened 2025-04-28
Plat UI 3629 poc service navigation

PatrykRudnicki:


play-frontend-hmrc-play-30/src/main/scala/uk/gov/hmrc/hmrcfrontend/views/Aliases.scala
```diff
@@ -30,8 +30,10 @@ trait Aliases {
   val CheckProgress = viewmodels.accountmenu.CheckProgress
 
   val Cy = viewmodels.language.Cy
+  val CyServiceNavigation = viewmodels.language.CyServiceNavigation
 
   val En = viewmodels.language.En
+  val EnServiceNavigation = viewmodels.language.EnServiceNavigation
```
PatrykRudnicki, line 36: Two additional models for language created to avoid changing what we have currently. The difference is in name. `ENG` instead of `English` and `CYM` instead of `Cymraeg`.

Probably as future implementation we can combine those into one model with 3 fields

play-frontend-hmrc-play-30/src/main/scala/uk/gov/hmrc/hmrcfrontend/views/viewmodels/header/Header.scala
```diff
@@ -62,6 +64,13 @@ case class Header(
       Some(LanguageToggle(defaults ++ x.linkMap))
     case None    => None
   }
+
+  val languageToggleServiceNav: Option[LanguageToggle] = inputLanguageToggle match {
+    case Some(x) =>
+      val defaults: SortedMap[Language, String] = SortedMap(EnServiceNavigation -> "", CyServiceNavigation -> "")
+      Some(LanguageToggle(defaults ++ x.linkMap))
+    case None    => None
+  }
```
PatrykRudnicki, line 73: We need to prepare a `languageToggle` for Service Navigation, so this part is duplicated too. Mostly because different default values. Again probably could be combined in one method

play-frontend-hmrc-play-30/src/main/twirl/uk/gov/hmrc/hmrcfrontend/views/components/HmrcHeader.scala.html
```diff
@@ -147,11 +174,6 @@
     @userResearchBanner.map { urb =>
         @hmrcUserResearchBanner(UserResearchBanner(language, urb.url))
     }
-    @if(displayHmrcBanner){
-      <div class="govuk-width-container">
-        @hmrcBanner(Banner(language, useTudorCrown))
-      </div>
-    }
```
PatrykRudnicki, line 154: We need to double check if we should be changing the order of banners. Currently based on the prototype HMRC banner is going on top

PR #354, opened 2025-06-09
PLATUI-3794: Refactored HmrcHeader to use new HeaderParams viewmodel

JoPintoPaul:


play-frontend-hmrc-play-30/src/main/scala/uk/gov/hmrc/hmrcfrontend/views/viewmodels/header/v2/HeaderParams.scala
```diff
@@ -0,0 +1,126 @@
...
+      signOutHref = headerParams.headerUrls.signOutHref,
+      inputLanguageToggle = headerParams.languageToggle,
+      userResearchBanner = headerParams.banners.userResearchBanner,
+      phaseBanner = headerParams.banners.phaseBanner,
+      additionalBannersBlock = headerParams.banners.additionalBannersBlock,
+      menuButtonLabel = headerParams.menuButtonOverrides.menuButtonLabel,
+      menuButtonText = headerParams.menuButtonOverrides.menuButtonText,
+      navigationLabel = headerParams.navigationOverrides.navigationLabel,
+      rebrand = headerParams.logoOverrides.rebrand
+    )
```
oscarduignan, line 118: do we need to have it both ways, or just Header => HeaderParams - with it going this way I think we'll have the chance to lose some data when we add new fields to HeaderParams

CHANGELOG.md
```diff
@@ -8,6 +8,22 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 For compatibility information see `govukFrontendVersion` and `hmrcFrontendVersion` in
 [LibDependencies](project/LibDependencies.scala)
 
+## [12.5.0] - 2025-06-012
```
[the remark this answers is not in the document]
JoPintoPaul: @TimothyFothergill thank you! Amended.

TimothyFothergill (dismissed):
nice

play-frontend-hmrc-play-30/src/main/scala/uk/gov/hmrc/hmrcfrontend/views/viewmodels/header/v2/HeaderParams.scala
```diff
@@ -0,0 +1,132 @@
...
+      headerUrls = HeaderUrls(
+        homepageUrl = header.homepageUrl,
+        serviceUrl = header.serviceUrl,
+        signOutHref = header.signOutHref,
+        assetsPath = header.assetsPath
+      ),
+      headerNames = HeaderNames(
+        productName = header.productName,
+        serviceName = header.serviceName
+      ),
```
oscarduignan, line 66: just thinking about the work to change for teams, is it worth having an entry in readme that maps the params to new locations?

but saying that, I can't remember how many people are directly using hmrcheader - might not be worth it if it's only a small number

one bit I thought of while considering that, is how we could reduce the work, not something for now, but if we have to do it again, maybe rather than nesting everyhing, just nest some stuff that's not commonly used, or group stuff more so there are less places to look (**edit:** squinting at it I think it's not adding anything it's just me doing equivalent of pushing pixels around the screen, so mostly just ignore this)

```
HeaderParams(
  ServiceDetails(
    urls & names
  )
  HeaderNavigation(
  )
  HeaderBanners(
    ...
  )
  HeaderDisplayOptions(
    classes & attributes & button text / label  & logo flags
  )
)
```

oscarduignan (dismissed):
I think only "worth doing anything?" any maybe not, is if we want to have anything in readme that explains the change - because I think sometimes changelog stuff is missed

oscarduignan (approved):

PR #367, opened 2025-09-04
PLATUI-3943: ServiceNavigation in HmrcHeader

JoPintoPaul:


play-frontend-hmrc-play-30/src/main/twirl/uk/gov/hmrc/hmrcfrontend/views/helpers/DeprecatedHmrcLayout.scala.html
oscarduignan, line 1: I think we could remove it from this file- I think this is only used by teams indirectly via https://github.com/hmrc/play-frontend-hmrc/blob/3b7a3da5aa00ed34d3d27020a5ae5271ebaa654e/play-frontend-hmrc-play-30/src/main/scala/uk/gov/hmrc/hmrcfrontend/views/html/helpers/HmrcLayout.scala#L28-L56

So adding it here won't make it usable for anyone

I think this indirection was to make people using HmrcLayout get a deprecation warning - which didn't seem like it was possible to attach directly to a generated twirl template so HmrcLayout was replaced with a class that delegated to twirl template that was renamed to DeprecatedHmrcLayout

PR #372, opened 2025-09-22
PLATUI-3958 implicit language convertion + ServiceNavigationSlot refactor

PatrykRudnicki:


oscarduignan:
looks good to me, only thing I think is that it needs a CHANGELOG.md entry - but probably omit it from README for now and maybe we can add that in later

PR #404, opened 2026-04-20
PLATUI-4266

oscarduignan:
Propagate the usage of service navigation into query params of links to common platform components like contact forms and accessibility statements

This is to address an accessibility issue - services using service nav linking to pages of these common platform components which don't use service nav yet are non compliant with wcag 2.2 rules around consistent identification due to the changing labelling of the language switch controls

And we can't enable service nav on those services without making all services not yet using service nav non-compliant

Although this is a patch release - some components now need a request header in scope which didn't before, and the shape of service nav config has changed - this should hopefully not require any changes for most use cases in hmrc services

To enable this behaviour, we needed a way to determine if service nav is being used so that our components can display the correct urls to common components

To do that, we've repurposed the feature flag we added to force use of service nav before migrating to v13 to enable people to migrate without any code changes

That feature flag was controlled only by config before, and that meant that services that needed to be able to conditionally display pages using service nav couldn't use it - but now it's an interface - which our components which need to change their output depending on use of service nav can depend on

It has a default config based implementation matching the previous behaviour, but that can be overridden by services which need more granular control with different strategies- we've included two generic strategies for different use cases - one where use of service nav is determined by a query parameter in the url and another where use of service nav is determined by a [play framework request attribute](https://www.playframework.com/documentation/3.0.x/Highlights26#Request-attributes)

Examples of how to use each will be detailed in guidance that will be included as part of this pull request

# Tasks

- [x] document how to change strategy with simple example, and what will be needed
- [x] update changelog
- [x] propagate service nav usage to tracking.js script tag via data attribute
- [x] propagate to standard phase banner urls
- [x] propagate to footer links
- [x] propagate to accessibility statement links 
- [x] propagate to report technical problem links

# Decisions

* should we propagate service nav usage to tracking.js via a data attribute?
  * probably should start passing it and tracking.js can also have a fallback- I guess the risk of not passing it is that the user journey could be using service nav but the page currently on has nothing on it to indicate that if it's not got welsh translation or service name - there's a need for consistency with pages earlier than current page in journey
* what query param to use?
  * `?useServiceNav` seems succinct and balanced to me, doesn't need to be understandable by users
* should query param be a boolean or based on presence
  * trade-off is that based on presence we can't fallback to config - but seems like when you are at the point you want to have it determined by config you can just switch to the by config strategy
* where should how to change service navigation config strategy be documented? We will have a blog post - it seems like having it in the repo hides it a bit and it might be easier to have it in a release note or something like that on github and link to that from blog post - makes it easy for us to tweak it there if needed adhoc - though I'm presuming it's only a small number who might refer to it
  * for now will document here, and copy that into release notes I guess - then presume to have a summary section on the blog post about what you can do if you're an embedded service - maybe as a link to a google doc

# Embedded services that want to display pages with and without service nav during migration

> [!IMPORTANT]
> **This guidance only applies to services using play-frontend-hmrc v12 - once you are using play-frontend-hmrc v13 it isn't possible to opt-out of the use of service navigation.**
>
> The conditional display of pages within a service using service nav and not - is something which should only be needed by embedded services which participate in user journeys initiated by other services - and only in the case where you can't coordinate the use of service nav across all those initiating services at the same time - and you want to allow them to update to use service navigation incrementally for a period of time.

Previously you would not use the service navigation feature flag in this case - instead you would just explicitly pass the service nav to our components yourself depending on when you wanted it to display

However, now that we have some deeply nested components that need to know if the service nav is being used - rather than making a lot of api changes to thread that context through components, we've turned the feature flag into an interface that you can swap out with another implementation if needed - and the only which needs to be present to determine it's usage is a request header which is already present in most cases.

## use service nav based on `play-frontend-hmrc.forceServiceNavigation` feature flag from config (default strategy)

This is how it behaved before, you don't need to do anything for it to continue to behave like this - if you were using this to enable service nav then once you update you won't likely won't need to do anything, and the necessary links to our common platform services like contact forms, accessibility statements, and global info pages like terms and conditions will now have an additional query param to tell those pages that they should be displayed with the service navigation.

## use service nav based on query param `?useServiceNav` 

to enable this you need to disable the default strategy and enable the module which configures this

in your application.conf
```
play.modules.disabled += "uk.gov.hmrc.hmrcfrontend.config.DefaultServiceNavConfigModule
play.modules.enabled += "uk.gov.hmrc.hmrcfrontend.config.ServiceNavCanBeControlledByQueryParamModule"
```

then if a request comes in with ?useServiceNav the page will be displayed using the service nav as if `play-frontend-hmrc.forceServiceNavigation` had been set, otherwise it will be displayed as normal.

if there are parts of your service that should change based on the use of service navigation then you would need to inject ServiceNavigationConfig and make your changes based on the forceServiceNavigation feature flag -which with this strategy in place will be determined by the presence of the query parameter only

if you have internal links through anchor tags and form actions to other pages in your service, then you will need to propagate the query param through those urls - to help with this, there is a helper `ServiceNavigationConfig.propagateViaQueryParam(url: String)` which can append the parameter to a url for you if service navigation is enabled

## use service nav based on a request attribute (or indicate to our components your use of it)

Request attributes allow you to store extra information inside request objects to be used later in the processing of a request. They can be accessed through the request header which is normally passed through views. We've added a strategy that checks for the presence and value of a request attribute to determine if the service nav is being used so that we can link to the correct versions of pages of common platform components.

in your application.conf
```
play.modules.disabled += "uk.gov.hmrc.hmrcfrontend.config.DefaultServiceNavConfigModule
play.modules.enabled += "uk.gov.hmrc.hmrcfrontend.config.ServiceNavCanBeControlledByRequestAttr"
```

then in your actions or filters in your service - to indicate that a request will be using or should use service navigation - you would set it on the request something like this if set in an action:

```
import uk.gov.hmrc.hmrcfrontend.config.ServiceNavCanBeControlledByRequestAttr.UseServiceNav

class ExampleController {

  def index(journeyId: String): Action[AnyContent] = Action.async { request =>
    Journey.findById(journeyId).map {
      case Some(journey) => {
        given Request[AnyContent] = request.addAttr(UseServiceNav, journey.useServiceNav)
        
        Ok(renderPage(journey))
      }
      case None =>
        NotFound("this is just an example to help you understand, please don't copy this code verbatim")
    }
  }

}
```

Where UseServiceNav should be set to true or false, and will fall back to the value of `play-frontend-hmrc.forceServiceNavigation` - and the example above gives a made up example of how you might map that from some state you have saved about the current journey - maybe from when it was initiated. You could also set this request attribute in a filter for all requests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,7 +26,7 @@ For compatibility information see `govukFrontendVersion` and `hmrcFrontendVersio
 
 ### Compatible with
 
-- [hmrc/hmrc-frontend v6.116.0](https://github.com/hmrc/hmrc-frontend/releases/tag/v6.111.0)
+- [hmrc/hmrc-frontend v6.116.0](https://github.com/hmrc/hmrc-frontend/releases/tag/v6.116.0)
 - [alphagov/govuk-frontend v5.14.0](https://github.com/alphagov/govuk-frontend/releases/tag/v5.14.0)
 
 
@@ -40,7 +40,7 @@ For compatibility information see `govukFrontendVersion` and `hmrcFrontendVersio
 
 ### Compatible with
 
-- [hmrc/hmrc-frontend v6.116.0](https://github.com/hmrc/hmrc-frontend/releases/tag/v6.111.0)
+- [hmrc/hmrc-frontend v6.116.0](https://github.com/hmrc/hmrc-frontend/releases/tag/v6.116.0)
 - [alphagov/govuk-frontend v5.14.0](https://github.com/alphagov/govuk-frontend/releases/tag/v5.14.0)
 
 ## [12.31.0] - 2026-02-04
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,29 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 For compatibility information see `govukFrontendVersion` and `hmrcFrontendVersion` in
 [LibDependencies](project/LibDependencies.scala)
 
-## [12.33.0] - 2026-02-26
+## [12.32.1] - 2026-04-21
+
+### Changed
+
+- propagate use of service nav via query params on links to common platform services
+- this is a patch to previous major version to introduce a fix for an accessibility issue
+- while it's a patch version it changes the api slightly of 3 things, which now require an 
+  implicit request header in scope:
+  - hmrc language select component
+  - hmrc report technical issue component
+  - service navigation config
+- service navigation config has been refactored to an interface and some alternative
+  strategies for determining if service navigation should be used have been provided
+  to assist "embedded" services that may need to conditionally serve pages with and
+  without service navigation during the migration period.
+
+### Compatible with
+
+- [hmrc/hmrc-frontend v6.116.0](https://github.com/hmrc/hmrc-frontend/releases/tag/v6.111.0)
+- [alphagov/govuk-frontend v5.14.0](https://github.com/alphagov/govuk-frontend/releases/tag/v5.14.0)
+
+
+## [12.32.0] - 2026-02-26
 
 ### Changed
 
@@ -21,7 +43,7 @@ For compatibility information see `govukFrontendVersion` and `hmrcFrontendVersio
 - [hmrc/hmrc-frontend v6.116.0](https://github.com/hmrc/hmrc-frontend/releases/tag/v6.111.0)
 - [alphagov/govuk-frontend v5.14.0](https://github.com/alphagov/govuk-frontend/releases/tag/v5.14.0)
 
-## [12.32.0] - 2026-02-04
+## [12.31.0] - 2026-02-04
 
 ### Changed
 
@@ -33,7 +55,7 @@ For compatibility information see `govukFrontendVersion` and `hmrcFrontendVersio
 - [alphagov/govuk-frontend v5.14.0](https://github.com/alphagov/govuk-frontend/releases/tag/v5.14.0)
 
 
-## [12.31.0] - 2026-02-04
+## [12.30.0] - 2026-02-04
 
 ### Changed
 
@@ -48,7 +70,7 @@ For compatibility information see `govukFrontendVersion` and `hmrcFrontendVersio
 - [hmrc/hmrc-frontend v6.111.0](https://github.com/hmrc/hmrc-frontend/releases/tag/v6.111.0)
 - [alphagov/govuk-frontend v5.14.0](https://github.com/alphagov/govuk-frontend/releases/tag/v5.14.0)
 
-## [12.30.0] - 2026-01-22
+## [12.29.0] - 2026-01-22
 
 ### Changed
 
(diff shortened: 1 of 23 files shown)


CHANGELOG.md
```diff
@@ -59,29 +81,19 @@ For compatibility information see `govukFrontendVersion` and `hmrcFrontendVersio
...
-
-### Changed
-
-- Added ADR relating to how we model optional properties in Scala viewmodels (documentation change only)
-
-### Compatible with
-
-- [hmrc/hmrc-frontend v6.107.0](https://github.com/hmrc/hmrc-frontend/releases/tag/v6.107.0)
-- [alphagov/govuk-frontend v5.13.0](https://github.com/alphagov/govuk-frontend/releases/tag/v5.13.0)
-
```
oscarduignan, line 84: because it was only a documentation change it didn't trigger a release

play-frontend-hmrc-play-30/src/test/scala/uk/gov/hmrc/hmrcfrontend/views/config/StandardPhaseBannerSpec.scala
oscarduignan, line 1: had to make some changes to these tests because I switched to get the config from the app container rather than build it myself

I am now strong proponent of injecting all the things xD think we still have a lot of stuff left over from past where play templates didn't have ability to have stuff injected into constructor

it-play-30/src/test/scala/uk/gov/hmrc/hmrcfrontend/views/components/HmrcReportTechnicalIssueIntegrationSpec.scala
oscarduignan, line 1: had to change these things because of type related stuff changing with new things being injected

play-frontend-hmrc-play-30/src/main/scala/uk/gov/hmrc/hmrcfrontend/config/ServiceNavigationConfig.scala
```diff
@@ -17,15 +17,87 @@
...
+trait ServiceNavigationConfig {
+  def forceServiceNavigation(implicit request: RequestHeader): Boolean
 
-  /*
-  Service Navigation is not hidden by the feature flag, but
-  we are adding this flag to allows enabling Service Navigation without any changes
-  Once this flag is set to true, we will add Service Navigation to the header, move service name and language toggle
-   */
-  val forceServiceNavigation: Boolean = config.get[Boolean]("play-frontend-hmrc.forceServiceNavigation")
+  def propagateViaQueryParam(url: String)(implicit request: RequestHeader): String =
```
JoPintoPaul, line 32:
```suggestion
  def bindQueryParam(url: String)(implicit request: RequestHeader): String =
```

Thoughts...?

play-frontend-hmrc-play-30/src/main/scala/uk/gov/hmrc/hmrcfrontend/config/ServiceNavigationConfig.scala
```diff
@@ -17,15 +17,87 @@
...
+
+  private def queryParamAlreadySetIn(queryString: String): Boolean =
+    queryString
+      .stripPrefix("?")
+      .split("&")
+      .exists(p => p == useServiceNavQueryParam || p.startsWith(s"$useServiceNavQueryParam="))
+}
+
+@Singleton
+class ServiceNavCanBeControlledByRequestAttr @Inject() (config: Configuration) extends ServiceNavigationConfig {
```
JoPintoPaul: Just so that I understand this:

* `ServiceNavCanBeControlledByRequestAttr` - this can take the useServiceNav from config or from request attribute?
* `ServiceNavCanBeControlledByQueryParam` - this can take the useServiceNav from query string param only?
* `ServiceNavCanBeControlledByConfig` - this can take the useServiceNav from config only?

JoPintoPaul:
@oscarduignan overall looks great. My only comments would be:

1. `ServiceNavCanBeControlledByRequestAttr` - I'd be interested to know the use case for this, as I wonder if there's a specific circumstance where we'd want to use this?
2. "ServiceNav..." vs. "ServiceNavigation" - discussed in Slack

Otherwise this looks all good!

JoPintoPaul (approved):

oscarduignan:
patch release has now been created from this 🙏

play-frontend-hmrc-play-30/src/main/scala/uk/gov/hmrc/hmrcfrontend/config/ServiceNavigationConfig.scala
oscarduignan: sorry jo I totally missed this and have only just seen it when sharing something with Tim 😮‍💨

PR #411, opened 2026-08-04
Plat UI 4386 govuk frontend upgrade

JoPintoPaul:


play-frontend-hmrc-play-30/src/main/scala/uk/gov/hmrc/govukfrontend/views/viewmodels/dateinput/DateInput.scala
```diff
@@ -43,6 +46,10 @@ case class DateInput(
   errorMessage: Option[ErrorMessage] = None,
   formGroup: FormGroup = FormGroup.empty,
   fieldset: Option[Fieldset] = None,
+  day: Option[InputItem] = None,
+  month: Option[InputItem] = None,
+  year: Option[InputItem] = None,
+  values: Map[String, Int] = Map.empty,
```
oscarduignan: should the value here should be a string

because we can accept months as words, like "oct"

oscarduignan: `Map[String, String]`

JoPintoPaul: This was me - originally this was `String` but examples failed as this was being passed through as integer from the JSON - I could make this into a forgiving reads where it can read either an int or string as `String`

play-frontend-hmrc-play-30/src/main/scala/uk/gov/hmrc/govukfrontend/views/viewmodels/dateinput/InputItem.scala
```diff
@@ -26,6 +26,7 @@ final case class InputItem(
   name: String = "",
   label: Option[String] = None,
   value: Option[String] = None,
+  error: Option[Boolean] = None,
```
oscarduignan, line 29: looking at the govuk template this could be reduced to `error = false` I think

there's one spot in it where it says `if error != false` but I think (though it's just as a squint) it doesn't actually make a difference

I'm not totally sure on this one though, so "it's complicated but necessary" is answer which needs no further explanation from my pov

JoPintoPaul: @oscarduignan I believe it's complicated but necessary - however I will double-check

play-frontend-hmrc-play-30/src/main/twirl/uk/gov/hmrc/govukfrontend/views/components/GovukButton.scala.html
```diff
@@ -18,15 +18,25 @@
...
+
+@buttonContent = @{
+    content match {
+        case htmlContent: HtmlContent if isStartButton && htmlContent.value.nonEmpty =>
+            Html("<span>" + content.asHtml + "</span>")
+        case _ =>
+            content.asHtml
+    }
+}
+
```
oscarduignan, line 30: will `Html("<span>" + Html() + "</span>")` work how we expect? concat string and html and wrap in another html? Could any double escape happen? We could use the twirl match

```suggestion
@buttonContent = {
    @content match {
        case NonEmptyHtml(_) if isStartButton => {<span>@content.asHtml</span>}
        case _                                => {@content.asHtml}
    }
}
```

not sure if that compiles - happy to be corrected if it is ok to append html and strings / won't be any double escaping

play-frontend-hmrc-play-30/src/main/scala/uk/gov/hmrc/govukfrontend/views/viewmodels/dateinput/InputItem.scala
JoPintoPaul: @oscarduignan - can confirm, modelling as `Option[Boolean]` is complicated but necessary for the behaviour where `false` and undefined are not equivalent in the tests where an error is defined but the `values` with error isn't explicit.

play-frontend-hmrc-play-30/src/main/twirl/uk/gov/hmrc/govukfrontend/views/components/GovukDateInput.scala.html
```diff
@@ -24,63 +24,99 @@
...
-            Label(
-              classes = "govuk-date-input__label",
-              content = item.label.mapNonEmpty(Text.apply).getOrElse(Text(item.name.nunjucksCapitalize))
-            ),
-          classes = s"govuk-date-input__input ${item.classes}",
-          autocomplete = item.autocomplete,
-          pattern = item.pattern,
-          attributes = item.attributes
-        ))
+      @{dateInputItem(item)}
```
oscarduignan: do we need the curly braces here? (asked to learn not request change)

play-frontend-hmrc-play-30/src/test/resources/fixtures/govuk-frontend/test-fixtures/button-start-link-with-html/output.txt
```diff
@@ -0,0 +1,6 @@
+<a href="/" role="button" draggable="false" class="govuk-button govuk-button--start" data-module="govuk-button">
+  <span>Start <em>now</em></span>
```
oscarduignan, line 2: my q of double encoding - I guess if this fixture is fine then it is fine!

play-frontend-hmrc-play-30/src/test/resources/fixtures/govuk-frontend/test-fixtures/date-input-day-and-month-(using-items)/output.txt
oscarduignan, line 1: I didn't realise we could use ( ) in filenames :TIL:

play-frontend-hmrc-play-30/src/main/twirl/uk/gov/hmrc/govukfrontend/views/components/GovukDateInput.scala.html
JoPintoPaul: @oscarduignan We do not! Removed now.

play-frontend-hmrc-play-30/src/main/scala/uk/gov/hmrc/govukfrontend/views/viewmodels/dateinput/DateInput.scala
JoPintoPaul: @oscarduignan Pushed up a change for this.

CHANGELOG.md
```diff
@@ -8,6 +8,21 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
...
 
+## [13.10.0] - 2026-08-06
+
+### Changed
+
+- Uplifted `govuk-frontend` to v6.4.0
+- Uplifted `hmrc-frontend` to v7.26.0
+- `GovukButton` component wraps Html content into <span> for start buttons
+- `GovukPanel` has new functionality to act as Interruption panel to pause user's journey and give important information
+- `GovukDateInput` has additional fields to customise individual items
```
oscarduignan:
```suggestion
- `GovukDateInput` has additional fields to customise individual items - and now automatically adds govuk-input--error class to date input items if date input has an errorMessage and items have no error class already
```

oscarduignan: not sure if I have worded that quite right - but thought it's worth calling out a change in behaviour - even if it won't impact people many probably / negatively

play-frontend-hmrc-play-30/src/main/scala/uk/gov/hmrc/govukfrontend/views/viewmodels/dateinput/InputItem.scala
oscarduignan: tbh it makes sense to me - I won't send you back to look at conditional stuff - so tricky to unpick and this is something people won't manually use most of the time

oscarduignan: sorry I must have not refreshed page

play-frontend-hmrc-play-30/src/main/twirl/uk/gov/hmrc/govukfrontend/views/components/GovukButton.scala.html
PatrykRudnicki: Refactored it to use `NonEmptyHtml` as that's basically what I was doing earlier, but cleaner. I kept Html(...) as version without it didn't work for me, but will try to change to it :)

oscarduignan: not needed because I found test case showing `Html("<span>" + Html() + "</span>")` seems to work!

oscarduignan: ah cool!

oscarduignan: sorry for some reason comments aren't loading for me immediately

CHANGELOG.md
PatrykRudnicki: Updated in separate commit as I didn't know if the suggestion would be with you as commit author :D If that would happen, we would need to ask Tim for approval :D

oscarduignan: legend

oscarduignan (approved):
Image

PR #415, opened 2026-09-15
PLATUI-4432: Uplift frontend dependencies

JoPintoPaul:


oscarduignan (approved):
looks good to me, should we put a comment inline in service navigation implementation that says "we have chosen not to implement align inline functionality for now due to some ongoing work on hmrc language select PLATUI-XXX" ? so for anyone coming later saying "why isn't this here" when they look in that file it should be clear that it's intentional omission

